Issue #54, opened 2014-12-05
Broken declaration in updateRouting() 


ksn135:
✗ app/console -V
Symfony version 2.6.2-DEV - app/dev/debug
✗ app/console cache:clear

  [Symfony\Component\Debug\Exception\ContextErrorException]  
  Runtime Notice: Declaration of Admingenerator\GeneratorBundle\Command\GenerateAdminAdminCommand::updateRouting() should be compatible with  
  Sensio\Bundle\GeneratorBundle\Command\GenerateBundleCommand::updateRouting(Sensio\Bundle\GeneratorBundle\Command\Helper\QuestionHelper $questionHelper, Symfony\Component\Console\Input\InputInterface $input, Symfony\Component\Console\Output\OutputInterface $output, $bundle, $format)

PR #55, opened 2014-12-05
Patch: remove dialog helper

ksn135:
Update commands to support QuestionHelper. See issue #54


loostro:
Created a branch based on tag beta.25 `pre-adminlte` and applied pr #55 to it.

Issue #73, opened 2014-12-17
Filtering on associated fields


loostro:
As reported by @calvera in [commit#2373398](https://github.com/symfony2admingenerator/GeneratorBundle/commit/237339872887bb0b46c5d20dc7d95d8049b7d125#commitcomment-9005815), the `getAllFields` does not return associations, which makes int improssible (with the current implementation) to filter by associations.

loostro:
@sescandell I think this can be closed, as #96 should fix it, right?

sescandell:
Yes it should

PR #74, opened 2014-12-17
quick fix for issue#73

calvera:
this is simple fix for #73


Builder/Admin/ListBuilder.php
```diff
@@ -49,7 +49,11 @@ public function getFilterColumns()
 
     protected function findFilterColumns()
     {
-        foreach ($this->getAllFields() as $columnName) {
+        $columns = $this->getVariable('filters');
+        if (null == $columns || 0 == sizeof($columns)) {
+            $columns = $this->getDisplayColumns();
```
sescandell: If we want to be able to filter on non displayed fields, you should better call `getAllFields()`

Could you please update this? And so we should be able to merge it.

Thank you!

loostro: :+1: for merge, when this is updated

sescandell: Finally not.

I'm working on this, and, I'm not sure this is enough.

I should push a PR including this change and some more. I think will can close that one

loostro:
Closeing. Will be fixed in another PR by @sescandell

PR #89, opened 2015-03-26
Move jquery to scripts_head block

loostro:
jQuery must be loaded in the head of the page, becouse some form widgets can use inline Javascript depending on jQuery.

Eg. `symfony2admingenerator/FormExtensionsBundle` has a lot `jQuery(document).ready(function { /* do smg here */ })` calls, which failed due to jQuery not being loaded first.

### DIFF
--- a/Resources/views/base.html.twig
+++ b/Resources/views/base.html.twig
@@ -29,7 +29,6 @@
 {% block javascripts %}
     {% javascripts package='admingenerator_js' filter='?uglifyjs2'
     output='assetic/js/compiled/admingenerator.min.js'
-        'admin/components/jquery/dist/jquery.js'
         'admin/components/bootstrap/dist/js/bootstrap.js'
         'admin/components/iCheck/icheck.js'
         'admin/components/slimScroll/jquery.slimscroll.js'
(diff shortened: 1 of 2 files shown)


loostro:
All of this becouse jQuery is undefined when

```html+django
jQuery(document).ready(function($) {
  // some js init code
});
```

is hit

ex-console log

sescandell:
In my own opinion it should not...

You can still update your base to make widgets including their inline script after jQuery (moving the `form_js` call after jquery)

loostro:
This would make useing inline jQuery scripts completely impossible.

For example, I want to customize some block, eg. to change button color on click. Normally, I could just add inline javascript block within `jQuery(document).ready()`. With Jquery not avaliable - I can't do that.

sescandell:
Of course you can.

You just have to do it the right way adding your script after the jQuery one.

loostro:
Not for inline scripts. When you modify, say `column_abc` block, and add 

```html+django
{{ parent() }}

<script type="text/javascript">
// here do some javascript, but only after the page is fully loaded
// so we use jQuery(document).ready()
// but if it's at the bottom of the page (javascripts block)
// then it will be undefined here
</script>
```

PR #100, opened 2015-03-31
Updated Dutch translations

bobvandevijver:
Reverted the translation for the Excel action, as the current is no Dutch. Also checked the other translations quickly.


loostro:

PR #101, opened 2015-03-31
Replace classify with method to handle dots

bobvandevijver:
The current implementation with `classify` does not work, as it does not translate dots in the name (which can be used to indicate relations). And as dots are not allowed in methods, it does not work. 

I've also added the capitalize filter such that the method still start with `formatSomething`. I would not use the classify filter anymore as it might result in conflicts with names.

### DIFF
--- a/Resources/templates/CommonAdmin/ExcelAction/ExcelBuilderAction.php.twig
+++ b/Resources/templates/CommonAdmin/ExcelAction/ExcelBuilderAction.php.twig
@@ -111,7 +111,7 @@ class ExcelController extends ListController
             $getter = 'get'.ucfirst($getter);
             $value = method_exists($data, $getter) ? $data->$getter() : null;
 
-            $formatedValue = $this->format{{ name|replace({".": "_"})|capitalize }}($value);
+            $formatedValue = $this->format{{ name|classify|php_name }}($value);
 
             // Convert DateTime object to given format
             if ($formatedValue instanceof \DateTime){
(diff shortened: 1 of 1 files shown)


loostro:
@bobvandevijver there does `classify` originate from? I can't find it in Twig and it's also not in our Twig Extensions?

bobvandevijver:
Classify was located in `\Doctrine\Common\Inflector\Inflector` (before in `Common\Util`), so you would be correct to not find it in Twig or the s2a extensions.

loostro:
No, I meant the filter. Nevermind - I found it. It was moved to [TwigGenerator](https://github.com/symfony2admingenerator/TwigGenerator)

loostro:
@bobvandevijver We have a filter for that already [php_name](https://github.com/symfony2admingenerator/TwigGenerator/blob/master/src/TwigGenerator/Extension/PHPPrintExtension.php#L58). Could you change the `|replace({".": "_"})|capitalize` to `|classify|php_name` ?

bobvandevijver:
@loostro I can not find anything about classify in the repo you've linked, but I found the declaration of the filter in [BaseBuilder](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Builder/BaseBuilder.php)

loostro:
@bobvandevijver ahh there it is :)

bobvandevijver:
@loostro Here you are!

loostro:

PR #103, opened 2015-04-01
Allow non standard db types

loostro:
Requested by @bobvandevijver on gitter. Use case scenario: if you have defined non standard database field types (like `phone_number` in his case) there was no way to set the default form/filter handler.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -80,115 +80,115 @@ class Configuration implements ConfigurationInterface
                     ->children()
                         ->arrayNode('doctrine_orm')
                             ->prototype('array')
-                                ->addDefaultsIfNotSet()
-                                ->children()
-                                    // datetime types
-                                    ->scalarNode('datetime')->defaultValue('datetime')->end()
-                                    ->scalarNode('vardatetime')->defaultValue('datetime')->end()
-                                    ->scalarNode('datetimetz')->defaultValue('datetime')->end()
-                                    ->scalarNode('date')->defaultValue('date')->end()
-                                    // time types
-                                    ->scalarNode('time')->defaultValue('time')->end()
-                                    // number types
-                                    ->scalarNode('decimal')->defaultValue('number')->end()
-                                    ->scalarNode('float')->defaultValue('number')->end()
-                                    // integer types
-                                    ->scalarNode('integer')->defaultValue('integer')->end()
-                                    ->scalarNode('bigint')->defaultValue('integer')->end()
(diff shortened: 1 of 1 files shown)


sescandell:
Hi there,

That's wrong, you can already make it without "new configuration".
You can simply define your own class thanks to `admingenerator.fieldguesser.doctrine.class` parameter (talking about the Doctrine ORM implementation, bt it works too with ODM and Propel) than override dedicated functions.

Here is a quick sample about what this new class might be:
```php
class DoctrineORMFieldGuesser extends BaseDoctrineORMFieldGuesser
{
 ...
    public function getFormType($dbType, $columnName)
    {
        static $myCustomTypes = array('phone_number' => 'text');
        
        return array_key_exists($dbType, $myCustomTypes) ? $mysCustomTypes[$dbType] : parent::getFormTypes($dbType, $columName);
    }
}
```

Or if you don't want to create your own, you can still make a compiler pass to add your configurations to the ones from  admingenerator (do you need a sample code for this?)

bobvandevijver:
@sescandell I already have a working which basically does what you imply, and were currently testing it. When ready, I make an PR against this branch.

sescandell:
My point is more: I'm not sure we need to add a specific configuration for this particular case. 

We can already easily make it just by overriding an existing parameter. I think we should better make a doc about that specific case than a dedicated development.

But that's just my opinion, if everybody else think we need this feature as a configuration... let's do it...

bobvandevijver:
I would definitely add this to this bundle: this eases the configuration in such extend that it would be weird to not support it (simply just adding three lines in your `config.yml` for an extra type, that's the nicest). Adding a complete compiler pass (or overwriting parts of the bundle) is in my opinion really not needed, and is also something prone to errors (and slowing down the kernel).

loostro:
Closeing as this code is wrong. I'll have a look at @tobias-93 PR.

sescandell:
Hi,

> Adding a complete compiler pass (or overwriting parts of the bundle) is in my opinion really not needed

Agree with you that adding a CompilerPass is over complicated "just for this". But creating your custom class using the "class parameter" dedicated for your tipical use case is not so painfull and is, in m own opinion, a good way to handle this specific use case. We are here talking about a very application specific use case. Otherwise, if you never use the "class parameter" for this usage... it simply means the parameter is useless.

I really think this issue is really specific and might be managed thanks to what is already existing.

About performances it has no impact as this is used only at initialization and because we are just talking about overriding a class (I don't talk about the compiler class which is not a good solution)

PR #105, opened 2015-04-02
Non standard dbtypes

loostro:


loostro:
TODO: add validate method which should load the default options if they are not present

loostro:
@sescandell Please review. The PR solves the problem of non standard fieldtypes and also adds a nice feature: you only have to define the field types you want to overwrite. 

Previous to this patch, if you wanted to change one field type, eg. `boolean => choice`, you had to copy/paste all the default types. Now you can simply do:

```
admingenerator_generator:
    formTypes:
        doctrine_orm:
            boolean: checkbox
    # no need to copy the default values for other field types
```

bobvandevijver:
@loostro @sescandell Exactly want I wanted to achieve :+1:

sescandell:
OK for me, nice one !

loostro:
@tobias-93 thanks for help :) merged

tobias-93:
@loostro cool, you're welcome!

PR #96, opened 2015-03-28
Filters features

sescandell:
Some various fixes/improvements on filters:
 - [x] Let's filters take into account groups
 - [x] Manage filters fields order
 - [x] Fix groups management
 - [x] Manage filters on association fields
  - [x] ORM
  - [x] ODM
  - [x] Propel
 - [x] Making filters credentials different than list fields ones
 - [x] Let scopes able to filter on secured fields

Fix issues: #82 , #73 , #58 , #36

PR #113, opened 2015-04-04
Allow filtering on associations by setting them filterable

tobias-93:
When adding the property 'filterable' to columns, they will only appear under filters if they are a field of the Entity.
When there's an association present with the property 'filterable', it will not appear under filters even though filtering works fine.
This is due to the behavior of the `getAllFields` method in the FieldGuessers, which only returns fields and not associations.
This PR fixes this.

### DIFF
--- a/Guesser/DoctrineORMFieldGuesser.php
+++ b/Guesser/DoctrineORMFieldGuesser.php
@@ -37,6 +37,7 @@ class DoctrineORMFieldGuesser extends ContainerAware
         if (!$this->doctrine->getManagerForClass(self::$current_class)->getConfiguration()->getMetadataDriverImpl()->isTransient($class)) {
             $this->metadata[self::$current_class] = $this->doctrine->getManagerForClass(self::$current_class)->getClassMetadata($class);
         }
+        
         return $this->metadata[self::$current_class];
     }
 
--- a/Guesser/DoctrineORMFieldGuesser.php
+++ b/Guesser/DoctrineORMFieldGuesser.php
@@ -37,13 +37,12 @@ class DoctrineORMFieldGuesser extends ContainerAware
         if (!$this->doctrine->getManagerForClass(self::$current_class)->getConfiguration()->getMetadataDriverImpl()->isTransient($class)) {
             $this->metadata[self::$current_class] = $this->doctrine->getManagerForClass(self::$current_class)->getClassMetadata($class);
         }
-
         return $this->metadata[self::$current_class];
     }
 
     public function getAllFields($class)
     {
-        return $this->getMetadatas($class)->getFieldNames();
+        return array_merge($this->getMetadatas($class)->getFieldNames(), $this->getMetadatas($class)->getAssociationNames());
     }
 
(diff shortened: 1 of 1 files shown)


tobias-93:
@loostro @sescandell It could be that I'm breaking desired behavior, in that case I'd prefer an extra method `getAllAssociations` which is then incorporated in the filters part. Also, the other fieldGuessers need to be updated as well if this fix is approved.

sescandell:
Hi @tobias-93 

I'm already working on a refactoring of filters and filtering on associations. Check #96 

It should handle your case

tobias-93:
Hi @sescandell, I know you're working on it but in your work I didn't see a fix for filtering on associations in the current filter usage (by setting filterable true). Could be that I'm missing it, but otherwise this is an easy and quick way to make all functionality available through the filterable property.

sescandell:
Here it is: https://github.com/symfony2admingenerator/GeneratorBundle/pull/96/files#diff-9da10e225fd96c3984b51a7a3920f74aR53

A typical config file is so:
```yml
builders:
    list:
        params:
            title: post.title.list
            display: [ author, title, tags, expirationDate, category ]
            # Customizing filters
            fields:
                isPublished:
                    addFormOptions:
                        required: false # Stupid sample, just to show how it works
            filters:
                - author
                - category
                - isPublished
            scopes:
```

But I can also include your fix for the specific case where the user doesn't add a filter param.

tobias-93:
I don't know about changes in doctrine but I suspect something like that. It would be great if you'd incorporate this in your filters work (which is great by the way!)

sescandell:
Do you have any idea about ODM and Propel how do they handle this?

tobias-93:
I don't use any of those, but I think I've made a fix for the ODM in #114. Propel is not my cup of tea in any sense ;)

sescandell:
OK I take it!

Thanks

loostro:
I'm :+1: for this. @sescandell do we merge this now, or we wait for your filters PR?

loostro:
Oh, and @tobias-93 could you add the same change to DoctrineODMFieldGuesser?
According to [this doc](http://www.doctrine-project.org/api/mongodb_odm/1.0/class-Doctrine.ODM.MongoDB.Mapping.ClassMetadataInfo.html#_getAssociationNames) DoctrineODM classmetadata also has the `getAssociationNames()` method.

loostro:
@tobias-93 @sescandell  I think the Propel ORM equivalent would be `getForeignKeys()` according to [this doc](http://propelorm.org/documentation/cookbook/runtime-introspection.html#tablemap-properties).

Usage would be in PropelORMFieldGueser `$this->getMetadatas()->getForeignKeys()`.@cedriclombardot could you have a look, I'm not sure

sescandell:
@loostro 

> I'm :+1: for this. @sescandell do we merge this now, or we wait for your filters PR?

I included these changes in PR #96

loostro:
OK, so we close this one and wait for #96 to be merged.

PR #114, opened 2015-04-04
Alternative solution to allow filtering on associations as proposed in #113

tobias-93:
As proposed in #113, it would be nice to have filtering on associations using the `filterable` property.
This PR does the same in a less rigorous way.

@loostro @sescandell Please review and discuss whether this PR or #113 is the best solution. Personally, I prefer this one as it doesn't influence the already existing `getAllFields`.

TODO: Propel, as I don't know enough about it to add `getAllAssociations` to it. It is not broken, but I'm not sure if they face the same issues as Doctrine users and definitely don't know how to fix it.

### DIFF
--- a/Builder/Admin/BaseBuilder.php
+++ b/Builder/Admin/BaseBuilder.php
@@ -237,13 +237,23 @@ class BaseBuilder extends GenericBaseBuilder
     }
 
     /**
-     * Retrieve all columns
+     * Retrieve all field columns
      *
      * @return array
      */
     protected function getAllFields()
     {
         return $this->getFieldGuesser()->getAllFields($this->getVariable('model'));
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


loostro:
The first PR (113) is much simpler. However, I do not use metadatas often. How is this solution better to the one proposed in #113 ?

tobias-93:
Now I don't change the getAllFields. Although I don't know whether it's used somewhere else, I think it's better to keep the current behavior in the fieldguesser and only change the filtering.

sescandell:
I think we can use your proposed solution in #113 

`getAllFields` is also used as a default value for `display` param for lists. So, I think it makes sense to have the same behavior in filters and columns by default.

So :-1: for that one but :+1: for #113

PR #122, opened 2015-04-09
Use PropertyAccessor

loostro:
Move the logic to traverse relation fields to `getValueForCell` and use PropertyAccessor.


sescandell:
:+1:

PR #124, opened 2015-04-10
Fix parameter key conflicts

loostro:
When useing queryFilter to add custom contraints, eg.:

```php
$queryFilter->addDefaultFilter('year', 2015);
$queryFilter->addDefaultFilter('budget', 'Q1');
```

with filters (on the same fields -> `year` and `budget` enabled), there are possible conflicts.

The custom contraint will create a `budget` param with value `Q1`.
The `budget` filter will create a `budget` param with value `%Q1%` (becouse it uses by default the `addStringFilter`) and will overwrite the previous `Q1` value.

This causes no results returned, becouse the query will be

```SQL
AND v0_.year = ? 
AND v0_.budget = ? 
AND v0_.budget LIKE ?
```

with params `[2015, '%C1%', '%C1%']`. This is wrong, as the parameters should be  `[2015, C1, '%C1%']`. The problem is becouse the first parameter (custom constraint) has the same name as the filter parameter and gets overwritten.

### DIFF
--- a/QueryFilter/BaseQueryFilter.php
+++ b/QueryFilter/BaseQueryFilter.php
@@ -6,23 +6,12 @@ abstract class BaseQueryFilter implements QueryFilterInterface
 {
     protected $query;
 
-    protected $paramNumber = 0;
-
     /**
      * @param $query
      */
     public function __construct($query)
     {
         $this->query = $query;
-    }
-
-    /**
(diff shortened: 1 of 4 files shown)


loostro:
Ready to review.

sescandell:
:+1:

PR #126, opened 2015-04-10
Fix error due to renamed variable in ExcelBuilder

tobias-93:
An error was introduced in #122, where the column variable was renamed to builder_column, but just at one place. This fixes the other place as well, preventing errors.


loostro:
@tobias-93 thanks, I've changed the name to shorter - just `column`

tobias-93:
Excellent, thanks!

PR #132, opened 2015-04-12
Escape filters div class

### LINKED ISSUE #121: Escape filters missing
At the line https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/EditTemplate/fieldsets.php.twig#L34 we print the formType in lowercase but we don't escape undesired characters, but we should.

Take a look to other templates to this particular situation too

loostro:
Fixes #121


sescandell:
:+1:

PR #134, opened 2015-04-12
Hotfix for recently merged PR

loostro:


Generator/DoctrineGenerator.php
```diff
@@ -42,10 +42,8 @@ protected function doBuild()
         );
         $generator->setFieldGuesser($this->getFieldGuesser());
         $generator->setMustOverwriteIfExists($this->needToOverwrite($generator));
-        if ($this->container->has('twig')) {
-            $generator->setTwigExtensions($this->container->get('twig')->getExtensions());
-            $generator->setTwigFilters($this->container->get('twig')->getFilters());
-        }
+        $generator->setTwigExtensions($this->twig->getExtensions());
```
loostro: @sescandell Is this actually needed? It's only present in Doctrine ORM generator. Could it be some legacy unused code? Or maybe Doctrine ODM / Propel generators are missing this part of code?

loostro:
Ahh wrong branch

Generator/DoctrineGenerator.php
sescandell: It should be necessary in all DBAL...

PR #141, opened 2015-04-15
Fix missing deps

loostro:


loostro:
Removed last container dependency

PR #142, opened 2015-04-15
FormOptions from config should override default options

matejvelikonja:
I think merging of this two array should be turned around, since otherwise you can't overide for example translation domain from generator config.

### DIFF
--- a/Resources/templates/CommonAdmin/EditType/type.php.twig
+++ b/Resources/templates/CommonAdmin/EditType/type.php.twig
@@ -46,10 +46,10 @@ class {{ builder.YamlKey|ucfirst }}Type extends BaseType
         $optionsClass = '{{ builder.namespacePrefixWithSubfolder }}\{{ bundle_name }}\Form\Type\{{ builder.BaseGeneratorName }}\Options';
         $options = class_exists($optionsClass) ? new $optionsClass() : null;
 
-        return $this->resolveOptions('{{ column.name }}', {{ column.formOptions|merge({
+        return $this->resolveOptions('{{ column.name }}', {{ ({
             'label': column.label,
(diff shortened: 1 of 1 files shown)


loostro:
It looks OK to me, but I want @sescandell to have a look too. Do you see any reason this shouldn't be merged? If no  I'm :+1: for a merge.

sescandell:
Hi,

As you can see here: https://github.com/matejvelikonja/GeneratorBundle/commit/3d3b53fff814368a035ad95ac67c9d9c058bafd3#diff-1f55542763294b75e96cb1b9a19be67a I liberally changed this order. I'm not sure, but if I remember well, this is because of translation_domain for checkbox fields.

Could you try, with your changes, that a boolean filter is well translated for example?

matejvelikonja:
@sescandell I don't really get it why the labels for form wouldn't be translated, this change only defines which options override each. If this change is not merged it is not possible to override translations_domain for only one form field.

sescandell:
I'm not saying this is the right way to do... I'm just saying that before merging your PR, we should test that choices fields are still translated otherwise we need to find a workaround for that

matejvelikonja:
@sescandell I've tried this dropdown with the config that you have sent me, and it works like it should. Se yes/no is translated.

screen shot 2015-04-17 at 15 25 56

sescandell:
What if in your generator, you define your own translation_domain thanks to something like:

```yml
# generator file
params:
    translation_domain: MyOwnTranslationCatalog
    ...
```

Thanks for your implication

matejvelikonja:
Yeah, I've tried to add my own translation domain and I'm also seeing translated filter as above.

sescandell:
OK, so good for me...

Thanks !

PR #143, opened 2015-04-16
[ConfigExtension] Fix fail on header.html.twig when no dashboard_route s...

cedriclombardot:
[ConfigExtension] Fix fail on header.html.twig when no dashboard_route specified = null


loostro:
@cedriclombardot merged, thanks

PR #144, opened 2015-04-16
Use Symfony URL generator for sort URLs

### LINKED ISSUE #140: Use symfony URL generator for sort URLs
At line [41](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ListTemplate/thead.php.twig#L41) and [43](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ListTemplate/thead.php.twig#L41) of `thead.php.twig` the sort URL is kinda hardcoded as `?sort={{sortOn}}&order_by={{order}}`, which is ugly and quite unhandy when you rely on the Symfony routing component as I do. 

I've implemented a tab-safe year-switcher, which works by adding a GET parameter to every URL on generation. So now, when clicking a sort, the default year is loaded instead of the year you were on (I hope you get what I mean :) )

So, instead of the wanted `?sort={{sortOn}}&order_by={{order}}&_year=2013`, I get the URL without the `_year` parameters. This can be solved by simply adjusting the [ListBuilderAction.php.twig](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ListAction/ListBuilderAction.php.twig#L56) such that the `getListUrl()` is splitted and calls the `getListRoute()` function to generate the URL. This function can then be reused in the `index.php.twig` to pass the route to the builder. 

I can implement a PR (but not today) if you agree with me, so let me know :)

bobvandevijver:
Fixes #140.

### DIFF
--- a/Resources/templates/CommonAdmin/ListAction/ListBuilderAction.php.twig
+++ b/Resources/templates/CommonAdmin/ListAction/ListBuilderAction.php.twig
@@ -55,7 +55,17 @@ class ListController extends BaseController
      */
     protected function getListUrl()
     {
-        return $this->generateUrl("{{ builder.routePrefixWithSubfolder ~ '_' ~ bundle_name ~ (builder.BaseGeneratorName ? '_' ~ builder.BaseGeneratorName : '') ~ '_list' }}");
+        return $this->generateUrl($this->getListRoute());
+    }
+
+    /**
+     * Get list index route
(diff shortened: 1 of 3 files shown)


loostro:

PR #145, opened 2015-04-19
Configurable excel tmp dir

loostro:


Resources/templates/CommonAdmin/ExcelAction/ExcelBuilderAction.php.twig
```diff
@@ -45,7 +45,7 @@ class ExcelController extends ListController
         $writer = $phpexcel->createWriter($phpExcelObject, '{{ builder.filetype }}');
         $response = new StreamedResponse(
             function () use ($writer) {
-                $tempFile = $this->get('kernel')->getRootDir().'/cache/' . 
+                $tempFile = '{{ builder.generator.bundleConfig.excel_tmp_dir }}/' .
```
matejvelikonja: is there a reason you didn't just use `$this->get('kernel')->getCacheDir()`?

loostro: @matejvelikonja no, no reason. actually that would be much simpler, I haven't thought of that :) I'll fix it :) thanks

matejvelikonja: np :)

loostro: @matejvelikonja this is what happens when you code till 2AM :P

PR #149, opened 2015-05-10
Allow generate admin into default AppBundle (without namespace)

ksn135:
Usage: app/console admin:generate-admin --namespace=AppBundle --dir=src --format=annotation


loostro:
@ksn135 merged, thanks

PR #152, opened 2015-05-19
Also set filters buttons to no display when defaultFiltersVisibility …

tobias-93:
…is collapsed

When using the defaultFiltersVisibility = 'collapsed', the buttons are still shown.
This commit fixes that.


loostro:
@tobias-93 merged, thanks

PR #156, opened 2015-06-01
Grid class can be set in entity generator

peric:
Grid class can be set in entity generator, which is then set in the view. If class is not set, default one (defined in template) is taken.

This can be used as workaround to quickly set custom grid class. Better solution would be that you can set default class globally (in config.yml), and also per field (maybe we can discuss this and find solution together). Order of priority would then be: Entity-generator.yml per field > Entity-generator.yml per builder > config.yml

In order to use this, you can define grid class under builders in generator file.
Example:

```yml
edit:
    params:
        title: "Custom title"
        display:
            - title
        actions:
            save: ~
            list: ~
        grid_class: "col-md-8"
```

Todo:
* I wanted to put this also in documentation, but didn't know where

### DIFF
--- a/Resources/templates/CommonAdmin/EditTemplate/fieldsets.php.twig
+++ b/Resources/templates/CommonAdmin/EditTemplate/fieldsets.php.twig
@@ -1,5 +1,6 @@
 {% block form_fieldsets %}
-    {% set gridClass = builder.variable('grid_class') %}
+    {% set gridClass = builder.variable('grid_class') | default('col-md-4') %}
+
     {% for fieldsetName,fieldset in builder.fieldsets %}
         {% set fieldsetColumns = builder.columnsFor(fieldset) %}
         {% set fieldsetGroups = fieldsetColumns|mapBy('groups')|flatten %}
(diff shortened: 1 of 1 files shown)


bobvandevijver:
I like this, but at the moment it removes the flexibility of setting the class per row. You can only have a global one, or the default one. So I guess that you would need to change that before this can be considered mergeable.

ksn135:
@peric Nice idea!
May be also add to new- and show- builders ?

peric:
@bobvandevijver that's right. Do you maybe have some example of how this can be configurable per field? 
I was trying to play with that, but didn't find any proper solution - saw just some examples with setting up that under tabs etc, but I didn't like it.

peric:
@bobvandevijver I've changed this so it won't break configuration per row.

@ksn135 you can use it in "new" and "edit". "show" uses different template.

matejvelikonja:
:+1: for merging

sescandell:
Hi,

> Do you maybe have some example of how this can be configurable per field? 

What about making it configurable per Column?

I think we can assume all fields would have the same width... Could you try to make it configurable per column.

All you need to do is:
 - define a new method `setGridClass` in the `Column` class
 - define a new method `getGridClass` in the `Column` class
 - change the template to use the `column.gridClass` property, and it should work

Thanks,

matejvelikonja:
actually I don't agree that all fields should be of the same width. For example you have two fields, one rich text field, where you write long text and this one should be wide. And you have for example ZIP code field which would look quite strange if it would be wide. So in my opinion it should be configurable per field.

sescandell:
Agree with @matejvelikonja (just mis-typed in my previous comment)

loostro:
I also agree that it has to be customizable per-field. The rich WYSIWYG editor vs simple number field is a very good example.

sescandell:
Thanks

PR #157, opened 2015-06-10
Fix route names in AppBundle without vendor

ksn135:
Update RoutingLoader.php

### DIFF
--- a/Routing/RoutingLoader.php
+++ b/Routing/RoutingLoader.php
@@ -139,6 +139,7 @@ class RoutingLoader extends FileLoader
                 }
 
                 $route = new Route($datas['pattern'], $datas['defaults'], $datas['requirements']);
+                $route_name = ltrim($route_name, '_'); // fix routes in AppBundle without vendor
                 $collection->add($route_name, $route);
                 $collection->addResource(new FileResource($controllerName));
             }


sescandell:
This PR breaks "everything".

You should also update templates which are still generating routes with underscores at first.
For example:
 - https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ListAction/ListBuilderAction.php.twig#L68
 - https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ListAction/ListBuilderAction.php.twig#L88

There is probably many other part of code like this.

@ksn135 could you please double check that?

Many thanks,

ksn135:
@sescandell sorry, PR is on the way...

ksn135:
Well, 33 matches across 18 files for string ```routePrefixWithSubfolder```
I believe that we need to do some refactoring and replace in templates many code like this ```{{ builder.routePrefixWithSubfolder }}_{{ bundle_name }}{{ builder.BaseGeneratorName ? "_" ~ builder.BaseGeneratorName : "" }}``` to simple function call, i.e. ```{{ builder.baseRoute }}```
@sescandell What do you think about it ?

sescandell:
That will centralize code, which is always a good thing.

Let's do this.

Thanks for your help,

ksn135:
@sescandell add PR #162

sescandell:
Merged.

I'll test it.

THanks,

PR #160, opened 2015-06-16
Allow filter options, for better customization of filter types

tobias-93:
I'd like to differentiate my filter options from my form options, as it makes not much sense to have the same options for the filter type as the form type.
This PR should not break BC, as form options is used if filter options is empty.

### DIFF
--- a/Builder/Admin/BaseBuilder.php
+++ b/Builder/Admin/BaseBuilder.php
@@ -192,6 +192,8 @@ class BaseBuilder extends GenericBaseBuilder
 
             if (array_key_exists('addFilterOptions', $fieldOptions)) {
                 $column->setAddFilterOptions($fieldOptions['addFilterOptions']);
+            } elseif (array_key_exists('addFormOptions', $fieldOptions)) {
+                $column->setAddFilterOptions($fieldOptions['addFormOptions']);
             }
         }
 
--- a/Builder/Admin/BaseBuilder.php
+++ b/Builder/Admin/BaseBuilder.php
@@ -189,6 +189,10 @@ class BaseBuilder extends GenericBaseBuilder
             if (array_key_exists('addFormOptions', $fieldOptions)) {
                 $column->setAddFormOptions($fieldOptions['addFormOptions']);
             }
+
+            if (array_key_exists('addFilterOptions', $fieldOptions)) {
+                $column->setAddFilterOptions($fieldOptions['addFilterOptions']);
+            }
         }
 
         return $column;
--- a/Builder/Admin/BaseBuilder.php
+++ b/Builder/Admin/BaseBuilder.php
@@ -151,14 +151,18 @@ class BaseBuilder extends GenericBaseBuilder
 
             if ($this->getYamlKey() === 'list') {
                 // Filters
-                $column->setFormOptions($this->getFieldOption(
+                $column->setFilterOptions($this->getFieldOption(
                     $column,
-                    'formOptions',
-                    $this->getFieldGuesser()->getFormOptions(
-                        $column->getFilterType(),
-                        $filteredFieldDbType,
-                        $this->getVariable('model'),
-                        $column->getFilterOn()
+                    'filterOptions',
+                    $this->getFieldOption(
+                        $column,
+                        'formOptions',
+                        $this->getFieldGuesser()->getFilterOptions(
+                            $column->getFilterType(),
+                            $filteredFieldDbType,
+                            $this->getVariable('model'),
+                            $column->getFilterOn()
+                        )
                     )
                 ));
             } else {
--- a/Generator/Column.php
+++ b/Generator/Column.php
@@ -89,6 +89,11 @@ class Column
     protected $formOptions = array();
 
     /**
+     * @var array
+     */
+    protected $filterOptions = array();
+
+    /**
      * @var string
      */
     protected $getter;
@@ -310,6 +315,16 @@ class Column
         return $this->filterType;
     }
 
+    public function setFilterOptions($filterOptions)
+    {
+        $this->filterOptions = $filterOptions;
+    }
+
+    public function getFilterOptions()
+    {
+        return $this->filterOptions;
+    }
+
     public function setLocalizedDateFormat($localizedDateFormat)
     {
         $this->localizedDateFormat = $localizedDateFormat;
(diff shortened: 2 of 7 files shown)


sescandell:
Actually this doesn't make any sense. You just have to think in another way: on List, a form has no meaning except for filters. So `formOptions` on lists are your equivalence to `filtersOptions`.
Knowing that, if you need to make some customization on your filter form, you can make them directly under the list builder in your generator file.

So, for me, this PR is not necessary

tobias-93:
@sescandell it makes as much sense as splitting up the types for form and filters. For a different type you want different options. If I use a certain type for my new and edit forms, I want to define the options in the 'general' part of the yml. If I then define other options in the list-builder, they get either merged or overridden by the general part, which can lead to errors as they may not be options at at the filter Type. Therefore, it seems logical to split the options as well as the types. Otherwise, I'd have to define my form options for new and edit in the respective builders doubly, making it less flexible...

bobvandevijver:
I agree with @tobias-93 that this feature is something that I would have been happy to see earlier. This makes it possible to define the different formType and filterType in one go, nicely together at the top of you config file, without having the need to overwrite something only under the filter section. 

So, although the PR is indeed not absolutely necessary as you say @sescandell, I do think that it is nice to have, and as it is not BC breaking, why not? 

However, I am missing the fact that it is not possible to overwrite the filterOptions with the formOptions given under the filter. But I haven't tested it, so I am not sure if this is a problem.

sescandell:
I see your point.

To be honest, in a recent refatoring I tried to introduced this "split".. but revert back my changes because I really thought `formOptions` are enough. I don't really see how adding a `filterOptions` might change things and make things more flexible. I'm not close to discussion... at the contrary, I'm glad to see ideas challenged. And if your proposition might made things easier... let's do it!

But let's more discuss about it first...

Here is how it is working now:
```yaml
generator: admingenerator.generator.doctrine
params:
    i18n_catalog: AdminSescandell
    model: Sescandell\AwesomeBundle\Entity\Question
    namespace_prefix: Sescandell
    concurrency_lock: ~
    bundle_name: AdminSescandellBundle
    pk_requirement: ~
    credentials: 'canManageAwesome(token)'
    fields:
        query:
            label: admin.awesome.label.query
            formType: text
            addFormOptions:
                required: true
                data_empty: ~
                widget: table 
...
builders:
    list:
        params:
            title: admin.awesome.title.list
            display:
                - query
                - activated
                - tags
            actions:
                new: ~
            object_actions:
                edit: ~
                copy: ~
                delete: ~
                toggleState: ~
            batch_actions:
                toggleState: ~
                addTags: ~
            fields:
                query:
                    addFormOptions:
                        required: false

```

New & Edit form have so the `query` field required. But not the filter. 
That said, all others options are still well configured. And, in my opinion, things are still easy to read and understand...

If we add the `filterOptions`, generators would have to look like this(let me know if I missed understood anything):
```yaml
generator: admingenerator.generator.doctrine
params:
    i18n_catalog: AdminSescandell
    model: Sescandell\AwesomeBundle\Entity\Question
    namespace_prefix: Sescandell
    concurrency_lock: ~
    bundle_name: AdminSescandellBundle
    pk_requirement: ~
    credentials: 'canManageAwesome(token)'
    fields:
        query:
            label: admin.awesome.label.query
            formType: text
            addFormOptions:
                required: true
                data_empty: ~
                widget: table 
            addFilterOptions:
                required:false
                data_empty: ~     # Copy / paste here
                widget: table        # Copy / paste here

...
builders:
    list:
        params:
            title: admin.awesome.title.list
            display:
                - query
                - activated
                - tags
            actions:
                new: ~
            object_actions:
                edit: ~
                copy: ~
                delete: ~
                toggleState: ~
            batch_actions:
                toggleState: ~
                addTags: ~
```

I don't see how this could be more flexible?



That said, appart of the discussion "does filterOptions might be interesting or not", the PR is incomplete. You also need to update the base builder class here for example: https://github.com/symfony2admingenerator/GeneratorBundle/blob/f488987fbc657eb9df994ddcc94c102216cbb432/Builder/Admin/BaseBuilder.php#L156 and double check tests https://github.com/symfony2admingenerator/GeneratorBundle/blob/9477510c145dc1c3cfeb2b97e18bd84b024f6ee1/Tests/Generator/ColumnTest.php

For information, during a recent refactoring on forms, I imagined to split formOptions and filterOptions... but ended with "formOptions are enough"... and I have to admit that was also because of code considerations to avoid copy paste part of code and/or configuration.

Differences between now and "after" are minor... but behind the scene, I think splitting both configurations will make things more complexe.

That said, if everyone is convinced except me... let's do it...

Waiting for your feedbacks, and thanks for your participation to all of this

tobias-93:
@sescandell thanks for the feedback.

This PR is not for when new, edit and filter use the same type, but rather for when the formType and filterType are different and the form/filtertype require different options. For example, we use a collection widget with configurable fields under the configurable_fields option for new and edit, while under the filters we use a search widget with a post_route and name_field option. If the configurable_fields option is used, the search type throws an error since it's not an option for that field. I agree that for most cases the same options (with different content) are ok, but in our case it sometimes isn't. In that case the filterOptions provide a nice way of working with that.

I've tried to fix your comments, but as I'm on a very slow internet connection right now I cannot test it exhaustively.

tobias-93:
Oh and I didn't really consider yet what would be different for formOptions or filterOptions in default, but I figured filters would never have to be required that's why I changed that one.

Builder/Admin/BaseBuilder.php
```diff
@@ -185,6 +189,12 @@ protected function createColumn($columnName, $withForms = false)
             if (array_key_exists('addFormOptions', $fieldOptions)) {
                 $column->setAddFormOptions($fieldOptions['addFormOptions']);
             }
+
+            if (array_key_exists('addFilterOptions', $fieldOptions)) {
+                $column->setAddFilterOptions($fieldOptions['addFilterOptions']);
+            } elseif (array_key_exists('addFormOptions', $fieldOptions)) {
```
sescandell: Why an `elseif` ? Couldn't we have both options?

tobias-93: @sescandell then the whole purpose of the differentiation between filter options and form options is defied, making the addition useless. Especially in the case of different types.

sescandell: Ok... I just miss-read the code.

Going to merge

sescandell:
Many thanks for that one ;)

PR #163, opened 2015-06-18
Support empty string filtering in scopes

tobias-93:
In the special case of wanting to filter on an empty string in scopes, this doesn't work due to the added % signs in the query filter.
This PR fixes that for Doctrine ORM.

### DIFF
--- a/QueryFilter/DoctrineQueryFilter.php
+++ b/QueryFilter/DoctrineQueryFilter.php
@@ -59,11 +59,16 @@ class DoctrineQueryFilter extends BaseQueryFilter
      */
     public function addStringFilter($field, $value)
     {
+
         list($tableAlias, $filteredField) = $this->addTablePathToField($field);
 
         $paramName = $this->getParamName($tableAlias.'_'.$filteredField);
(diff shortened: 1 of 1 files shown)


sescandell:
Hi @tobias-93 

If I well understand this is for handling generators like:

```yaml
builders:
    list:
        params:
            title: post.title.list
            display: [ author, title, tags, expirationDate ]
            filters:
                - author
                - isPublished
            scopes:
                "First Group":
                    "Empty":
                        filters: ~
                    "Unamed John":
                        default: true
                        filters:
                            "author": "John"
                            "name": ""
```

Is that right?

In that case, shouldn't the line ` $this->query->andWhere(sprintf('%s.%s LIKE :%s', $tableAlias, $filteredField, $paramName));` would have to be changed for something like ` $this->query->andWhere(sprintf('%s.%s = :%s', $tableAlias, $filteredField, $paramName));` ? (in that particular case of course).

Thanks

tobias-93:
@sescandell You're understanding it right. I've checked my PR in this configuration in our project, and it works like a charm with the `LIKE` configuration. I understand why the `=` would be better, but it isn't needed.

sescandell:
Ok

sescandell:
Thanks

PR #164, opened 2015-06-19
Updated FieldGuesser for boolean/checkbox combination

bobvandevijver:
I encountered the following problem:

When using a boolean field, which is not nullable in the database, the field guesser decides that the 
field is required, as it is not nullable.

However, as it is rendered as checkbox, the field should not be required, as then the browser forces it to be checked before a submit can be done.

I guess the best solution (if I am not missing anything) is to add another case to the Guesser, which detects a boolean and a checkbox form type. It should then just return the array('required' => false). That is what this PR does, for every manager (ORM, ODM and Propel).

### DIFF
--- a/Guesser/DoctrineODMFieldGuesser.php
+++ b/Guesser/DoctrineODMFieldGuesser.php
@@ -178,6 +178,13 @@ class DoctrineODMFieldGuesser extends ContainerAware
                 'translation_domain' => 'Admingenerator'
             );
         }
+        
+        if ('boolean' == $dbType &&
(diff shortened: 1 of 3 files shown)


sescandell:
Thanks

matejvelikonja:
There are some cases when checkbox is required. Is this still possible with this solution?

bobvandevijver:
@matejvelikonja The only thing this changed is the default guessed `required: true` to `required: false`. You can still set the required option in your configuration as usual, and thus overwriting the default guessed value.

PR #165, opened 2015-06-20
[BC Break] Update methods visibility

sescandell:
This PR doesn't fix any issue, it's just a "fix" to make methods `protected` whereas `public` where it doesn't make sense to be public.

I prefer to make a PR so everyone is aware of the BC Break introduced by these changes.

Let me know if you disagree with this PR


ksn135:
:+1: yes, please
https://insight.sensiolabs.com always mark it as error ))

tobias-93:
:+1: from me too

PR #166, opened 2015-06-20
[BC Break] Remove deprecated calls and services

### LINKED ISSUE #72: Deprecated the security.context service
It's NOT deprecated right now, but may be we should prepare for this BC early, if we already based on version 2.6. I'm talking about many setSecurityContext function calls in code that will become obsolete.
Quote:
```
// Symfony 2.5
$user = $this->get('security.context')->getToken()->getUser();
// Symfony 2.6
$user = $this->get('security.token_storage')->getToken()->getUser();

// Symfony 2.5
if (false === $this->get('security.context')->isGranted('ROLE_ADMIN')) { ... }
// Symfony 2.6
if (false === $this->get('security.authorization_checker')->isGranted('ROLE_ADMIN')) { ... }
```
Source: http://symfony.com/blog/new-in-symfony-2-6-security-component-improvements

sescandell:
## Fix 

Fix issue #72.

## Changes

 - Bumped Symfony version
 - `security.context` => `security.authorization_checker`
 - ~~`request` => `getRequest()`~~
 - `request` => method injection + protected property
 - `bind()` => `handleRequest()`
 - `form.csrf_provider` => `isCsrfTokenValid()`
 - Provide file content to `Yaml::parse` instead of file path
 - Route method is not anymore based on `requirements` option
 - `setDefaultOptions` => `configureOptions` 
 - update `setAllowedTypes` arguments call

If you think about other deprecated call, just let me know or make PR on that PR :)

## BC Break: 
 - Symfony version is fixed to 2.7 min
 - All actions (index, create, update, object, batch, filter, scope) now take as a first parameter the `Request` object
 - Generated forms method `setSecurityContext` have been renamed `setAuthorizationChecker`

### DIFF
--- a/Builder/Generator.php
+++ b/Builder/Generator.php
@@ -59,7 +59,7 @@ class Generator extends TwigGeneratorGenerator
     public function __construct($cacheDir, $yaml)
     {
         parent::__construct($cacheDir);
-        $this->setYamlConfig(Yaml::parse($yaml));
+        $this->setYamlConfig(Yaml::parse(file_get_contents($yaml)));
     }
 
     /**
@@ -354,7 +354,7 @@ class Generator extends TwigGeneratorGenerator
     protected function setYamlConfig(array $yaml)
     {
         $this->yaml = array_replace_recursive(
-            Yaml::parse(__DIR__.'/../Resources/config/default.yml'),
+            Yaml::parse(file_get_contents(__DIR__.'/../Resources/config/default.yml')),
             $yaml
         );
     }
--- a/CacheWarmer/GeneratorCacheWarmer.php
+++ b/CacheWarmer/GeneratorCacheWarmer.php
@@ -78,7 +78,7 @@ class GeneratorCacheWarmer implements CacheWarmerInterface
 
     protected function parseYaml($file)
     {
-        $this->yaml_datas = Yaml::parse($file);
+        $this->yaml_datas = Yaml::parse(file_get_contents($file));
     }
 
 }
--- a/Controller/Doctrine/BaseController.php
+++ b/Controller/Doctrine/BaseController.php
@@ -3,6 +3,7 @@
 namespace Admingenerator\GeneratorBundle\Controller\Doctrine;
 
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
 
 /**
  * A base controller for Doctrine
@@ -12,4 +13,8 @@ use Symfony\Bundle\FrameworkBundle\Controller\Controller;
  */
 abstract class BaseController extends Controller
 {
+    /**
+     * @var Request
+     */
+    protected $request;
 }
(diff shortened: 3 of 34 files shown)


sescandell:
Just see that `getRequest()` is deprecated too... going to update that...

tobias-93:
@sescandell I've found the following list when analyzing the deprecated log of our project:
* If we're upgrading to 2.7 anyway, why not change the signature of the setDefaultOptions in the form types to configureOptions as per 2.7?
* passing filenames to `Yaml::parse()` was deprecated in 2.2, this should be refactored to `Yaml::parse(file_get_contents($path))` (used in `Generator`, `GeneratorCacheWarmer`, `ControllerListener`, `RoutingLoader`, `BaseValidator`)
* OptionsResolver::setAllowedTypes method with an array of options is deprecated since version 2.6 and will be removed in 3.0. Use the new signature with a single option instead. (used in `BaseType`)
* The "_method" requirement is deprecated since version 2.2 and will be removed in 3.0. Use the setMethods() method instead. (used in `RoutingLoader`)
* The "pattern" option in file "/var/www/myfile.yml" is deprecated since version 2.2 and will be removed in 3.0. Use the "path" option in the route definition instead. (used in `NestedRoutingLoader` and `RoutingLoader`)

I could make a PR for these, but it will not be before Monday

sescandell:
Thank @tobias-93 

I'll check your list

sescandell:
@tobias-93 done!

tobias-93:
@sescandell Good work!

sescandell:
Merge delayed: using the request as a property object in controller is not a good thing at all.

bobvandevijver:
@sescandell Huh, what are you talking about?

bobvandevijver:
O, I see what you did now. But why is this not a good thing at all? Could you explain?

sescandell:
Because if you made a subrequest from a view your initial request object might be changed and so, when coming back to your "master" request, the request property object might be out-dated and invalid...

I don't really have a "real" example, but it might be possible and so complicated or confusing to analyze.

Here is fake sample:

```php
class MyControler extends AdminGenController
{
    public function myAction1(Request $request)
    {
        $this->request = $request;
        
        $renderedView = $this->renderView('mytemplate.html.twig');
        // Here $this->request is invalid because it contains the sub-request for myAction2
        return doSomethingWithRequest($renderView);
    }

    public function myAction2(Request $request)
    {
        $this->request = $request;
        // ...
    }
}
``` 

```twig
{# mytemplate.html.twig #}
<h1>Awesome</h1>
{{ render(controler('MyControler:myAction2')) }}
...
```

bobvandevijver:
@sescandell Okay, I see you point, but I guess it would be another instance of the controller. And secondly, when you are already in you view, it should not matter that the data in your controller is changed, as you will no longer use it. 

An other solution is to pass the Request object to every function, but that would be a big BC break. 
But, maybe the best is to also use the `request_stack`? With `$this->get("request_stack")->getCurrentRequest()` on the places where the request is needed?

sescandell:
I was more thinking about using `request_stack`!

sescandell:
@bobvandevijver actually, regarding at [ControllerResolver](https://github.com/symfony/HttpKernel/blob/master/Controller/ControllerResolver.php) class from Symfony Core, you're right, controller would be a new instance. So current implementation might work in this situation.

Let's keep it as this if you're agree.

If nobody have any cons about this, I'll merge it.

bobvandevijver:
I agree with this merge :+1:

PR #171, opened 2015-06-30
Fixes filtering on Many-to-many relations

tobias-93:
On Many-to-many relations, filtering was impossible. This PR fixes that for Doctrine ORM.
@sescandell feedback please


sescandell:
Thanks,

PR #173, opened 2015-07-08
Fix PropelORMFieldGuesser

Lionel09:
- fix bad attributes function call for ```getColumn```
- add spaces for indentation PSR compatible


sescandell:
Thanks

Issue #115, opened 2015-04-06
Show/hide filters panel


loostro:
Moved out of #96

Add a "show/hide" button to slide in/out filters panel. By default they should be hidden, to make "more room" (width) for more table columns.

Maybe(?) we could make the "default" behaviour configurable in generator.yml?

sescandell:
> Maybe(?) we could make the "default" behaviour configurable in generator.yml?

This is already provided in PR #96 

Otherwise :+1: for this enhancement

loostro:
in my recent project `filtersMode: top` the "collapse/expand" button does not work
@sescandell could you check in one of yours? is it broken only for me or for everyone?

sescandell:
I'll do, just have to update one project

sescandell:
Hi @loostro 

I just use the master branch on the demo project and collapsed/expand button on filter box is working well:
![expanded](https://cloud.githubusercontent.com/assets/1559970/7240666/1b51819a-e7b3-11e4-8dcf-f9ea2d8ffeed.png)

![collapsed](https://cloud.githubusercontent.com/assets/1559970/7240669/204b19f4-e7b3-11e4-8e8f-900cc27e6031.png)


I can push the demo version with my custom samples if you want (including Groups and filters customizations).

loostro:
@sescandell Thanks. So it seems I must have forgotten to add some class or include some JS when doing custom theme.

PR #181, opened 2015-07-24
Fix form tag position in FiltersBuilderTemplate.php.twig

ksn135:
Fix issue #115. Now in ```filtersMode: top``` show/hide button works as expected.


sescandell:
Merged, thanks

Issue #148, opened 2015-05-06
Format number choice


pjam:
In list view, when only 1 object is created, it reads "Showing all 1 results", where it should be "Showing all 1 result" or "Showing 1 result".

loostro:
This can be changed by editing the `nbresults.php.twig` template [here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ListTemplate/nbresults.php.twig#L12) adding another if 

(if results > 1  ---> current behaviour, if results == 1 --> new message)

sescandell:
The right way to do is to use a `trans_choice` in place of `trans`

pjam:
Yeah, that was my first thought, but looking at the code I see I would have to create the echo_transchoice method and for me its not worth the trouble, I'll go for the condition loostro suggested.
Anyway the issue is here for any contributor to take it and solve it.

pjam:
Is it possible to override the nbresults template without changing the vendor file? I've been trying to do it without success...

bobvandevijver:
@pjam Yes, but you will have to follow this cookbook item: [Extending generator templates](https://github.com/symfony2admingenerator/AdmingeneratorGeneratorBundle/blob/master/Resources/doc/cookbook/extending-generator-templates.md).

pjam:
Thanks @bobvandevijver exactly what I needed!

PR #192, opened 2015-08-07
Fix for nbresults.html.twig when there is only one result #148

pjam:
Added text for the result count in list view when it is showing only one result

Added english and portuguese text


bobvandevijver:
@pjam Can you try to add all translations? I will not merge this if not all translations are added. 

Here you have the dutch translation:
`<b role="results-count">%count%</b> resultaat.`

pjam:
I've added all the translations... I did my best.. ;)

bobvandevijver:
Merged, thanks!

PR #193, opened 2015-08-07
Fix to allow rendering a controller in the show template if no partial added

pjam:


bobvandevijver:
@pjam Can you resubmit this PR? It also contains the commit which is already submitted in #192

PR #194, opened 2015-08-11
Fixes Reset button being kept disabled when using modal filters

tobias-93:
When using the modal filters functionality, the reset button is disabled even when filters are used. This was due to the namespace being assimilated wrongly, and this PR fixes that.


bobvandevijver:
Merged, thanks!

PR #195, opened 2015-08-12
Update paginator.php.twig

calvera:
I think it is better to use dropup instead of dropdown. In case list is at bottom it displays list correctly...


sescandell:
Hi @calvera 

Thanks for your contribution.

Could you please attach a screenshot of this PR?

Thanks!

calvera:
Here it is...

display_n_elements_popup

bobvandevijver:
I agree with this! :+1:

sescandell:
Merged,

Thanks,

PR #197, opened 2015-08-13
Fix to allow rendering a controller in the show template if no partial added

pjam:
When trying to render a controller in show action, an error was thrown if not added any partials


sescandell:
Thanks,

PR #201, opened 2015-08-17
Update filters.md

enkaho:
Add missing option filterable: 1


bobvandevijver:
I do not agree with the documentation: I do not use the filterable option anywhere, and it works perfectly fine. Hence, it would not be a must, but more something you can try when troubleshooting.

enkaho:
you're right... it's ok after update.

PR #202, opened 2015-08-22
[PATCH] Fix filters update via GET request

ksn135:
When you set filters via GET request  in query string you can only set primary id of related object, NOT the object itself. I'm adding conditional check for such case.


sescandell:
Thanks

PR #204, opened 2015-08-26
[PATCH] Add omitted parameter to echo_if_granted calls

ksn135:
Without this patch no object instance passed to custom security function as described in https://github.com/symfony2admingenerator/AdmingeneratorGeneratorBundle/blob/master/Resources/doc/security/credentials.md
```yaml
# ...
           actions:
                list: ~
                edit: 
                    credentials:        'isEditableByUser(object)'   
# ...
```
PS: Please IGNORE PR #203


bobvandevijver:
Merged, thanks!

PR #218, opened 2015-12-13
Update dependencies

tobias-93:
KnpMenuBundle and SensioGeneratorBundle have been updated, but we cannot use the new versions as they're blocked by the composer.json of this bundle.
Please allow the new version.


loostro:
@tobias-93 merged thanks

Issue #217, opened 2015-12-09
Cache warmer issue w. Propel


Lionel09:
I'm facing a problem when using GeneratorBundle w. Propel ORM (1.6).

This problem occurs when you make change in services, or simply by removing the cache ```rm -fr app/cache``` for example. After that, the **first** call of command line like ```app/console cache:clear``` or the **first** load of a website page fails.

If I redo a ```app/console cache:clear``` or reload the website page after this initial issue, it works fine. But a ```composer update``` for example will always fail at the end of its process, so it's a very annoying bug.

Stack trace:

```
~/www/project.test$ app/console cache:clear --env=dev -vvv


                                                                                         
  [PropelException]                                                                      
  No connection information in your runtime configuration file for datasource [default]  
                                                                                         


Exception trace:
 () at /home/lionel/www/project.test/vendor/propel/propel1/runtime/lib/Propel.php:583
 Propel::getMasterConnection() at /home/lionel/www/project.test/vendor/propel/propel1/runtime/lib/Propel.php:561
 Propel::getConnection() at /home/lionel/www/project.test/vendor/propel/propel-bundle/Propel/PropelBundle/Security/Acl/MutableAclProvider.php:55
 Propel\PropelBundle\Security\Acl\MutableAclProvider->__construct() at /home/lionel/www/project.test/app/cache/dev/appDevDebugProjectContainer.php:2608
 appDevDebugProjectContainer->getSecurity_Acl_ProviderService() at /home/lionel/www/project.test/app/bootstrap.php.cache:2155
 Symfony\Component\DependencyInjection\Container->get() at /home/lionel/www/project.test/app/cache/dev/appDevDebugProjectContainer.php:2552
 appDevDebugProjectContainer->getSecurity_Access_DecisionManagerService() at /home/lionel/www/project.test/app/bootstrap.php.cache:2155
 Symfony\Component\DependencyInjection\Container->get() at /home/lionel/www/project.test/app/cache/dev/appDevDebugProjectContainer.php:2647
 appDevDebugProjectContainer->getSecurity_AuthorizationCheckerService() at /home/lionel/www/project.test/app/bootstrap.php.cache:2155
 Symfony\Component\DependencyInjection\Container->get() at /home/lionel/www/project.test/app/cache/dev/appDevDebugProjectContainer.php:3779
 appDevDebugProjectContainer->getTwigService() at /home/lionel/www/project.test/app/bootstrap.php.cache:2155
[...]

PR #219, opened 2015-12-18
Add a Propel date range filter & force Propel boot init

Lionel09:
cf https://github.com/symfony2admingenerator/FormExtensionsBundle/issues/52

It has to be improved:
- only for Propel
- need to make a function inheritage in the ListController to work

cf https://github.com/symfony2admingenerator/GeneratorBundle/issues/217

Check if Propel is used and force Propel::init before warmup

Note: sorry because I've failed to do 2 differents PR for this 2 issues

### DIFF
--- a/CacheWarmer/GeneratorCacheWarmer.php
+++ b/CacheWarmer/GeneratorCacheWarmer.php
@@ -39,12 +39,18 @@ class GeneratorCacheWarmer implements CacheWarmerInterface
      */
     public function warmUp($cacheDir)
     {
+        if ($this->container->has('admingenerator.generator.propel')) {
+            $this->propelInit();
+        }
+
         foreach ($this->finder->findAllGeneratorYamls() as $yaml) {
             try {
                 $this->buildFromYaml($yaml);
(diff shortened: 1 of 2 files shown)


sescandell:
Hi @Lionel09 

Thanks for your activities these days, we really appreciate it!

Speaking about the `addDateRangeFilter` I'm just not convinced about that modification. I think you should change the `addDateFilter` rather than creating a new one that is not referenced anywhere (meaning that except by overriding the controller, this function is never used). As far as I understand your update could be moved to the `addDateFilter`.

Does it make sense for your?

Lionel09:
yes it does, i'll update asap

sescandell:
Thanks, merged!

Lionel09:
Can you put theses PR updates in the Packagist repo too, to be able to make a "composer require"?

Sorry but I don't know exactly what is the process in this case.

sescandell:
Done: https://github.com/symfony2admingenerator/GeneratorBundle/releases/tag/v2.0-beta42

Just wait for some minutes so packagist is up to date

PR #221, opened 2015-12-27
Use credentials back instead of groups

### LINKED ISSUE #82: credentials in field context
Hi!

Actually I want to use credentials in field context:
```
  fields:
    myField:
      credentials: 'hasRole("ROLE_USER")'
```

After implementation of

```
params:
    fields:
        username:
            credentials: 'hasRole("ROLE_ADMIN")'
            label:            admin.user.fields.username

```
 every try gives an error:

```
Twig_Error_Runtime in classes.php line 4696: 
An exception has been thrown during the rendering of a template 
("Could not set option "credentials" on "username" column in "User" list builder.")
 in "ListBuilderAction.php.twig" at line 10.
```

After research, it seems to me, that field driven usage of credentials not is implemented. 
Do I am wrong?

If not, I like to think ahead:

Groups can be assigned to a field. So my idea would be to define groups in configuration file like:

```
params:
    groups:
        Admin:
            credentials: 'hasRole("ROLE_ADMIN")'
        User:
            credentials: 'hasRole("ROLE_USER")'
```
with usage

```
params:
    fields:
        username:
            groups:           ["User", "Admin"]
            label:            admin.user.fields.username
```

So in this case we need to change the getGroups method on EditController/NewController/ShowController like this:
``` EditBuilderAction.php.twig ```
```
    /**
     * Get groups.
     *
     * @param \{{ model }} ${{ builder.ModelClass }}
     * @return array
     */
    protected function getGroups(\{{ model }} ${{ builder.ModelClass }})
    {

        {% if true == builder.getVariable("groups") %}
        //ensure that it is not empty
        $groups = array("GROUPS_ENABLED");
            {% for key, group in builder.getVariable("groups") %}
                {% if group.credentials is defined %}

        if ($this->evaluate('{{group.credentials}}', ${{ builder.ModelClass }})) {
            $groups[] = "{{key}}";
        }

                {% else %}

        $groups[] = "{{key}}";

                {% endif %}
            {% endfor %}
        {% else %}
        $groups = array();
        {% endif %}

        return $groups;
    }
```

This would produce:
``` ~\EditController.php ```
```
    protected function getGroups(UserBundle\Entity\User $User)
    { 
        $groups = array("GROUPS_ENABLED");
                            
        if ($this->evaluate('hasRole("ROLE_ADMIN")', $User)) {
            $groups[] = "Admin";
        }
                                           
        if ($this->evaluate('hasRole("ROLE_USER")', $User)) {
            $groups[] = "User";
        }
                                    
        return $groups;
    }

```

What do you think about that?

### LINKED ISSUE #154: [Credentials] credentials and groups
Hi.

Could you give some pointers on how to use credentials in generator.yml?

I want a list field visible only with some credentials and tried
```yaml
fields:
    fieldname:
        credentials: [ROLE_SUPER_ADMIN]
```

without luck, error thrown "Could not set option "credentials" on "fieldname" column in "Entity" list builder.". Is it possible to use credentials for fields? How?

Regards

### LINKED ISSUE #167: [Credentials] Templates don't display fields if one requires a group that current user doesn't match
I just faced a limitation to how we manage credentials in templates.

If I want to make one field available only if you have a given group, I'll do this:
```yaml
generator: admingenerator.generator.doctrine
params:
    model:              Admingenerator\DoctrineOrmDemoBundle\Entity\Post
    namespace_prefix:   Admingenerator
    i18n_catalog:       AdmingeneratorOrmDemoBundle
    concurrency_lock:   ~
    bundle_name:        DoctrineOrmDemoBundle
    pk_requirement: ~
    fields:
        author:
            filterable: 1
            groups: [non]
            filtersGroups: [oui]
```

Except that, because of this, fields in the same fieldset of this "protected" field are not anymore displayed in views if you don't have the group of the one particular protected field.
This is because templates contain the following code:

```twig
{% set fieldsetColumns = builder.columnsFor(fieldset) %}
            {% set fieldsetGroups = fieldsetColumns|mapBy('groups')|flatten %}

            {% if fieldsetGroups is not empty %}
                {{ echo_set('fieldsetGroups', echo_twig_arr(fieldsetGroups), false) }}
                {{ echo_if('fieldsetGroups|intersect(groups) is not empty') }}
            {% endif %}
```

Meaning that, if one field in your set of fields requires a group and the current user doesn't have it, all the fieldset is not displayed.

Issue is not so for linked to groups, but more to how tests on fieldsets are performed.

A simple way to see this in action is to define a group property on a field and let the `display: ~` configuration in any builder: no field is so displayed.

sescandell:
After thinking about credentials, and groups functionnality (and in particulary issues related to groups), I think we should better back to a simple solution but as powerfull as groups.

The main idea is to remove "groups" terms and come back to credentials and use Symfony native capacities to handle security. With credentials (based on security authorization checker service), we can manage security in a "natural" way in a Symfony project using simple ROLE_XXX checks. For improved tests, users might still implement custom voters (which is re-usable).

JMS Extra Security Bundle has been moved as a suggested package. To enable it, use the new config parameter `use_jms_security` and and the bundle to your composer.json file.

This PR is a starting point for discussion as already started in issue #154 
What do you think about reverting back to simple credentials ?

I really think, we can do everything with credentials keeping things simple for simple needs, but still powerfull and flexible for advanced use cases thanks to Voters or JMSSecurityBundle.

Tasks:
 - [x] Developments
 - [x] Fix #167 
 - [x] Fix #154 
 - [x] Fix #82 
 - [x] JMS Security Bundle no more mandatory: use the parameter `use_jms_security`
 - [x] Fix commands to be generator-bundle compatible (or agnostic from it?)
 - [x] Removed deprecated calls to prepare to Symfony2.8 and 3.0 (Form ones not handle in this PR)
 - [x] Tests
 - [x] Write documentation
 - [x] Approve the PR


bobvandevijver:
I can only say :+1 for this! I did not test the PR, but I find current the group functionality unclear, while indeed JMS provides enough functionality for this. Good work!

tobias-93:
:+1: from my side, but this should be incorporated very well in the documentation including descriptions how to work with voters.

Builder/Generator.php
```diff
@@ -118,44 +118,50 @@ public function addBuilder(BuilderInterface $builder)
...
+            }
+
+            // Grab builder configuration only if defined
+            if (is_null($builder[$param])) {
+                continue;
+            }
+
+            $configurations = array();
+            foreach ($builder[$param] as $name => $configuration) {
+                if (!(is_array($configuration) || is_null($configuration))) {
```
loostro: Maybe `if (!is_array($configuration) && !is_null($configuration)) {` would be more clear whats going on here

.travis.yml
```diff
@@ -4,9 +4,7 @@ php:
     - 5.3
     - 5.4
     - 5.5
```
loostro: Maybe add PHP 5.6 and 7.0?

Resources/templates/CommonAdmin/EditAction/EditBuilderAction.php.twig
```diff
@@ -28,8 +27,7 @@ class EditController extends BaseController
     protected function getEditForm(\{{ model }} ${{ builder.ModelClass }})
     {
         return $this->createForm($this->getEditType(), ${{ builder.ModelClass }}, array_merge(
-            $this->getFormOptions(${{ builder.ModelClass }}),
-            array('groups' => $this->getGroups(${{ builder.ModelClass }}))
+            $this->getFormOptions(${{ builder.ModelClass }})
```
loostro: We should drop `array_merge` here.

Twig/Extension/ArrayExtension.php
```diff
@@ -41,6 +42,7 @@ public function mapBy(array $input, $key)
                 return $item[$key];
             }
 
+            // TODO: use PropertyAccessor ??
```
loostro: Why use PropertyAccessor here?

Twig/Extension/SecurityExtension.php
```diff
@@ -0,0 +1,74 @@
...
+        return array(
+            'is_one_admingenerator_granted' => new \Twig_SimpleFunction('is_one_admingenerator_granted', array($this, 'isOneGranted')),
+        );
+    }
+
+    /**
+     * @param  array  $credentials
+     * @return bool
+     */
+    public function isOneGranted(array $credentials, $object = null)
```
loostro: I really like this change. Previous implementation was more cumbersome. And since this is defined as a service, it can be easily overriden/tweaked to implement custom behaviour. Great job!

loostro:
I like this change, as it simplifies implementation, but keeps all the functionality. Great job @sescandell  :+1:

.travis.yml
sescandell: Added in #22cf90b

Thanks,

Twig/Extension/ArrayExtension.php
sescandell: It's not mandatory, just a proposition.
If I well understand, we are here doing like the "PropertyAccessor", why don't we use it?

sescandell:
Thanks :+1: 

And thank you for your comments & feedbacks

sescandell:
Ready to merge.

Tested on the demo repo.
Once merges, I'll push my modifications on that repo too.

loostro:
:+1: for merge

bobvandevijver:
I'm missing some upgrade notes/documentation. I'm working on new docs right now (#223), so maybe it is handy to add those to the PR?

loostro:
@bobvandevijver Lets move this discussion to #223 as this issue is closed now and #223 deals with documentation.

PR #224, opened 2016-01-06
Symfony 2.8 & 3.0 compatibility

sescandell:
Ready to review

Developments:

 - [x] Form references as FQCN
 - [x] Embed types
 - [x] Tests

### Initialization

The bundle initialization in you `AppKernel` file now need the kernel object as a parameter of the constructor.

### Embed types

`embed_types` in generators is now not anymore necessary. Use the FQCN form name of your generated forms, and everything will be fine!


sescandell:
Note : one deprecated notice comes from `knp-menu` and is depending on commit https://github.com/KnpLabs/KnpMenu/commit/18253a4467d60614735321387f250b0bc7b0ecad

No tag exists today

bobvandevijver:
@sescandell I changed some indentations in the list to keep them readable :)

sescandell:
> From @sescandell
 > Note : one deprecated notice comes from knp-menu and is depending on commit KnpLabs/KnpMenu@18253a4
 >
 > No tag exists today

I found a way to fix that on our side, I'll hand it in next commit

sescandell:
ping @loostro , @tobias-93 , @bobvandevijver

sescandell:
[Demo bundle](https://github.com/symfony2admingenerator/symfony2-admingenerator-demo-edition) is sync with that branch

tobias-93:
@sescandell I'm :+1: for merge (maybe a little more upgrading documentation would be welcome, but all right)

bobvandevijver:
@tobias-93 tested the admingenerator in our project, so I'm also :+1:

loostro:
@sescandell if you're waiting for my review, I will be able to do it on Friday

Form/BaseType.php
```diff
@@ -49,21 +41,6 @@ protected function checkCredentials($credentials, $model)
...
-     */
-    protected function inject($formType)
-    {
-        if (is_object($formType) && method_exists($formType, 'setAuthorizationChecker')) {
-            $formType->setAuthorizationChecker($this->authorizationChecker);
-        }
-
-        return $formType;
-    }
-
```
loostro: @sescandell i really like that auto-registering as service and removeing this part of code, it was kinda hacky

Generator/DoctrineGenerator.php
```diff
@@ -102,52 +95,4 @@ protected function doBuild()
             )
         );
     }
-
```
loostro: @sescandell how do we handle embed generators? is this feature removed?

Guesser/PropelORMFieldGuesser.php
```diff
@@ -279,26 +280,32 @@ protected function getOptions($type, $dbType, $model, $fieldPath, $filter = fals
...
         if ((\PropelColumnTypes::BOOLEAN == $dbType || \PropelColumnTypes::BOOLEAN_EMU == $dbType) &&
-            (preg_match("#^choice#i", $type) || preg_match("#choice$#i", $type))) {
-            return array(
+            preg_match("#ChoiceType$#i", $type)) {
+            $options = array(
                 'choices' => array(
-                   0 => 'boolean.no',
-                   1 => 'boolean.yes'
+                    'boolean.no' => 0,
+                    'boolean.yes' => 1
```
loostro: @sescandell current choice type implementation has inverted key - value in  `choices` option?

Resources/templates/CommonAdmin/ListAction/filters.php.twig
```diff
@@ -48,9 +56,76 @@
     {% endblock %}
 
     {% block getFilters -%}
+   /**
+    * @return array
+    **/
```
loostro: typo `**/` -> `*/`

Resources/templates/CommonAdmin/ListAction/filters.php.twig
```diff
@@ -1,4 +1,12 @@
 {% block filters %}
+
+   /**
+    * Action called on filter request. After the call, the user is redirected
+    * to the list page.
+    *
+    * @param Request $request The request
+    * @return RedirectResponse
+    **/
```
loostro: typo `**/` -> `*/`

loostro:
I'm curious about why embed generators feature is removed - otherswise :+1: for merge.

tobias-93:
@sescandell Why did you remove the convert_as_form filter? It breaks the workflow when using query builders in Forms. Or is there a different way to do it? That should then be in the documentation!

sescandell:
@loostro : I removed `embed_types` because, given my tests, they are not necessary anymore. You can simply set the Form to the generated one with its FQCN and it works (given that your generator is correct too). Did I missed a use case?

@tobias-93 : do you have any sample of a generator using this functionnality? I'm used to use form overriding for that (in PHP). But, if you're used to use `generators` (meaning some people might use it too), I'll revert that part (but if you have a code to share, I'm interested :))

Thanks,

tobias-93:
@sescandell we have something like:
<pre>
params:
  fields:
    mentors:
      formType: Admingenerator\FormExtensionsBundle\Form\Type\DoubleListEntityType
      addFormOptions:
        query_builder: function ($er) { return $er->createQueryBuilder("m")->join("m.person", "p")->orderBy("p.fullname"); }
        multiple: true
</pre>

loostro:
@sescandell yes, the use case (problem) embed types were solveing was when Generator A wanted to use a form generated by Generator B (eg. as nested form). Thus, if empty cache, it required to first generate Generator B (only the form types). If it is handled now automatically - then thats a very nice improvement.

tobias-93:
Ok my example is a bit lame (it can be solved using a sortOn), but you get the thing I want to do I hope?

sescandell:
@tobias-93 OK, I'll set `convert_as_form` back...

@loostro I'll push a sample in the demo project so we can  be sure it is working well

sescandell:
Except for `query_builder` does anyone use `convert_as_form` capacities?

@loostro @bobvandevijver @tobias-93 do you mind if I set as "deprecated" the `convert_as_form` function?
I personnaly never use this option. For these cases, I prefer to use `'getMyPropertyOptions()` method than writing PHP in YML files.

calvera:
@sescandell 

i'm not sure if it is convert_as_form use:

```yaml
        _start_range:
            label: Date Range
            filterable: true
            filterType: s2a_daterange_picker
            dbType: date
            formOptions:
                opens: left
                parentEl: div.modal
                ranges:
                    'Next 7 Days': "[moment(), moment().add(6, 'days')]"
                    'Tomorrow': "[moment().add(1, 'days'), moment().add(1, 'days')]"
                    'Today': "[moment(), moment()]"
                    'Yesterday': "[moment().subtract(1, 'days'), moment().subtract(1, 'days')]"
                    'Last 7 Days': "[moment().subtract(6, 'days'), moment()]"
                    'Last 30 Days': "[moment().subtract(29, 'days'), moment()]"
                    'This Month': "[moment().startOf('month'), moment().endOf('month')]"
                    'Last Month': "[moment().subtract(1, 'month').startOf('month'), moment().subtract(1, 'month').endOf('month')]"
```

sescandell:
@calvera after PHP inside YML, Javascript! I love to see how other people uses the bundle, we always found some things we never thought about :)

Actually, it should not be a `convert_as_form` option. I will check this is still working with this PR.

Thanks for sharing!

tobias-93:
@sescandell I'm not against deprecating stuff that's either hard to maintain or superseded by better options, but I'm just curious why this needs to be removed.
Furthermore in this case I can see that it can work in another way. My only problem would now be that I would have to adjust it in the `New`, `Filters` and `EditType`. The old behavior enabled us to keep everything the same for all Types (although that is not always needed, I can understand).
But I can live with it either way. Could you describe it in the Upgrade documentation?

sescandell:
@tobias-93 

No, thanks to @loostro , in the new generator, an `MyModelOptions` object can be used to avoid duplications of custom options in New, Edit and Filters form (no sample for now, but [available here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/f29b802f13c69c92b3e1806666e9cb0ecabd2004/Generator/BundleGenerator.php#L174) and [here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/skeleton/bundle/DefaultOptions.php.twig).)

We are all agree documentation need improvements. I know @bobvandevijver is working on it. I'll, on my side, work on the demo project too.

I've made a mistake removing it. I'll revert this modification (updating how it works, and adding tests).

Hopefully, you're all here to check my errors! :)

tobias-93:
@sescandell thanks!
We'll always be there ;)

bobvandevijver:
Wauw, I did not know about that MyModelOptions class... Will add it to my todo list!

Guesser/PropelORMFieldGuesser.php
sescandell: Yes, in Symfony2.8+ they inverted choices handling: http://symfony.com/doc/current/reference/forms/types/choice.html#choices

tobias-93: @sescandell I think you then also have to add `choices_as_values = true`...

sescandell: That is already the case... only if Symfony < 3.0 : https://github.com/symfony2admingenerator/GeneratorBundle/pull/224/files#diff-6258b29d64f9109f6ee13de74a5bcd25R247

sescandell:
I reverted back `convert_as_form` option. Just made some modifications and added some tests.

Is everything OK for you now?

Guesser/PropelORMFieldGuesser.php
tobias-93: Ah, I missed it.

tobias-93:
@sescandell could you add it to the form types again? This way it cannot be used yet...

sescandell:
Ohh, sorry, it seems I partially commited my changes...

I'll push it sunday (not on my personnal computer right now and I'll get it back only Sunday... sorry :/)

tobias-93:
@sescandell no problem. Next week I'll be on holiday so I'll review on the 1st of February. And I must say I never checked Symfony 3.0 so I cannot say anything about the compatibility with that (can only check 2.8 due to low PHP version :(

sescandell:
Ready for review (again :) )

Generator/DoctrineGenerator.php
sescandell: Actually, I see a difference compared to before. In current implementation, for embed types to work, Edit and New Controller need to be created (whereas they are not necessary).I will so bring back embedTypes.

sescandell: Issue created #247 , I'll make a PR

PR #225, opened 2016-01-06
Fix the use of expressions in the credentials handling

tobias-93:
When reverting the usage of groups to credentials, a few references to isGranted were missed, while a perfect implementation in the SecurityExtension was available. This PR fixes this discrepancy, so that also the formtypes and all templates use the isOneGranted method.

Without, a mixed behavior is yielded, which sometimes checks using the JMSSecurityExtraBundle structure and sometimes using the built-in isGranted, making it impossible to use the JMSSecurityExtraBundle structure.

### DIFF
--- a/Resources/templates/CommonAdmin/EditType/type.php.twig
+++ b/Resources/templates/CommonAdmin/EditType/type.php.twig
@@ -5,6 +5,9 @@ namespace Admingenerated\{{ namespace_prefix }}{{ bundle_name }}\Form\Base{{ bui
 
 use Admingenerator\GeneratorBundle\Form\BaseType;
 use Admingenerator\GeneratorBundle\Form\BaseOptions;
+{% if(admingenerator_config('use_jms_security')) %}
+use JMS\SecurityExtraBundle\Security\Authorization\Expression\Expression;
+{% endif %}
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
@@ -71,7 +74,12 @@ class {{ builder.YamlKey|ucfirst }}Type extends BaseType
     protected function canDisplay{{ column.name|classify|php_name }}(\{{ model }} ${{ builder.ModelClass }} = null)
     {
         {% if column.credentials is not empty %}
-            return $this->checkCredentials('{{ column.credentials }}', ${{ builder.ModelClass }});
+            {% if(admingenerator_config('use_jms_security')) %}
+                $credentials = new Expression('{{ column.credentials }}');
+            {% else %}
+                $credentials = '{{ column.credentials }}';
+            {% endif %}
+            return $this->checkCredentials($credentials, ${{ builder.ModelClass }});
         {% else %}
             return true;
         {% endif %}
--- a/Resources/templates/CommonAdmin/FiltersType/type.php.twig
+++ b/Resources/templates/CommonAdmin/FiltersType/type.php.twig
@@ -5,8 +5,11 @@ namespace Admingenerated\{{ namespace_prefix }}{{ bundle_name }}\Form\Base{{ bui
 
 use Admingenerator\GeneratorBundle\Form\BaseType;
 use Admingenerator\GeneratorBundle\Form\BaseOptions;
-use Symfony\Component\Form\FormBuilderInterface;
+{% if(admingenerator_config('use_jms_security')) %}
 use JMS\SecurityExtraBundle\Security\Authorization\Expression\Expression;
+{% endif %}
+use Symfony\Component\Form\FormBuilderInterface;
+
 
 class FiltersType extends BaseType
 {
@@ -57,7 +60,12 @@ class FiltersType extends BaseType
         protected function canDisplay{{ column.name|classify|php_name }}()
         {
             {% if column.filtersCredentials is not empty %}
-                return $this->checkCredentials('{{ column.filtersCredentials }}', null);
+                {% if(admingenerator_config('use_jms_security')) %}
+                    $credentials = new Expression('{{ column.filtersCredentials }}');
+                {% else %}
+                    $credentials = '{{ column.filtersCredentials }}';
+                {% endif %}
+                return $this->checkCredentials($credentials, null);
             {% else %}
                 return true;
             {% endif %}


bobvandevijver:
Seems ok to me, @sescandell can you check and merge?

sescandell:
Hi,

That's not ok for me.
There is actually two things.

1. I missed "use JMS Security Expression" option in forms => it seems to be true, and we should fix that => :+1: (but I think we should better update Forms templates so classes are generated with the right code [like in controllers](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/security_action.php.twig#L18) for example)
1. Changes I'm not ok with are ones belonging to tests on credentials as iterable. `is_one_granted` is not here to test "_an array of credentials to check if one of them is true_". Maybe I should have renamed `credentials` to `credential` but `credentials` should not be an array. It should be a valid string for `isGranted` method ([as mentionned here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/doc/security/credentials.md#2-usage-and-configuration)). The purpose of `is_one_granted` is to handle displaying fieldset depending on different column credentials inside that fieldset (or more globally any container). It should not be used to check a credential. Do you see what I mean?

The reason is, I don't see any case where we would say "that field should be display if the user has ROLE_A or ROLE_B". If you are in that situation, a best practice would be:
 1. To use JMS Expression if you really want to do this that way, and so specifying `hasRole(ROLE_A) OR hasRole(ROLE_B)` as `credentials` value
 1. To use a custom Voter (which is, IMO, a better way to handle that because you would create a semantic control that woul be centralized and reliable from anywhere in your application. What if tomorrow you need to move to ROLE_A or ROLE_B or (ROLE_C and isMorning) ?)


Let me know if all of this doesn't make sense for you,

Thanks,

tobias-93:
@sescandell Ok, I fixed your comments.
One thing: Why does the `isOneGranted` method expect an array and then iterate over it when it is not the expected behavior?

sescandell:
Sorry, this is probably because of my "poor" english, you misunderstood me :)

Let me try another time.

##### If JMSSecurity is not enabled
`credentials` parameter from generators must be a __string__. It will be used as the first parameter for call to the method `isGranted`. So `credentials` __must be__ a string.

`is_one_admingen_granted` is a specific Twig function for (and only for) display purposes. It is used in order to know if a block have to be displayed. We use that one for form fields (and other situations, but form is the simplest one). Imagine the following (simplified) generator:
```yml
...
fields:
    author:
        credentials: ROLE_MEMBER
    title:
        credentials: ROLE_MEMBER
    summary:
        credentials: ROLE_FOR_ALL_USERS
list:
    params:
        filters:
            display: [[author, title], [summary]]
``` 

Imagine the connected user doesn't have the `ROLE_MEMBER`. `is_one_admingen_granted` will prevent rendering the first empty fieldset that would normally contain the `author` and `title` fields. For that, we actually compute all credentials from all fields inside that fieldset, and check if at least one is valid. That's why it takes an array. But it doesn't check a field specifically. The "generated" code will looks like (schematically):
```php
$fieldsCredentials = ['ROLE_MEMBER', 'ROLE_MEMBER']; // Actually, when generated, values are unique
if (isOneAdmingenGranted($fieldsCredentials)) { //  false
    renderFieldsetContainingAuthorAndTitle(); // not called
}

$fieldsCredentials = ['ROLE_FOR_ALL_USERS'];
if (isOneAdmingenGranted($fieldsCredentials)) { // true
    renderFieldsetContainingSummary(); // called
}

...

function renderFieldsetContainingAuthorAndTitle() {
    print('<fieldset>');
    if (isGranted('ROLE_MEMBER') {
        print('Author');
    }

    if (isGranted('ROLE_MEMBER') {
        print('Title');
    }
    print('</fieldset>');
}

function renderFieldsetContainingSummary() {
    print('<fieldset>');
    if (isGranted('ROLE_FOR_ALL_USERS') {
        print('Summary');
    }
    print('</fieldset>');
}
```

Imagine now if the `display` parameter is `display: [[author, title, summary]]` (all fields in the same fieldset). In that situation, the fieldset would be displayed (because ROLE_FOR_ALL_USERS is valid), and then only the `summary` field will be rendered (because calls to `isGranted('ROLE_MEMBER')` will fail).
The "generated" code will looks like (schematically):
```php
$fieldsCredentials = ['ROLE_MEMBER', 'ROLE_MEMBER', 'ROLE_FOR_ALL_USERS']; // Actually, when generated, values are unique
if (isOneAdmingenGranted($fieldsCredentials)) { //  true, because of ROLE_FOR_ALL_USERS
    renderFieldsetContainingAuthorAndTitleAndSummary(); // called
}

 ... 

function renderFieldsetContainingAuthorAndTitleAndSummary() {
    print('<fieldset>');
    if (isGranted('ROLE_MEMBER') {
        print('Author');
    }

    if (isGranted('ROLE_MEMBER') {
        print('Title');
    }

    if (isGranted('ROLE_FOR_ALL_USERS') {
        print('Summary');
    }
    print('</fieldset>');
}
```

Thanks to this, we prevent having an empty `<fieldset></fieldset>` in the DOM. `is_one_admingen_granted` is only for display purposes.

So, for this, regarding `is_granted` and `is_one_admingen_granted`: everything should already be fine.

##### If JMSSecurity is enabled

Actually: it doesn't change anything in the logic. Just, before using the `credentials` parameter in a `isGranted` call, we simply wrapp it in an `JMS\Expression`.

All remaining things stay the same.

My error here, is that I forgot to take into account that configuration on Forms code generation.
We should update [that part of code](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/EditType/type.php.twig#L74) to become this:
```twig
    protected function canDisplay{{ column.name|classify|php_name }}(\{{ model }} ${{ builder.ModelClass }} = null)
    {
        {% if column.credentials is not empty %}
            {% if(admingenerator_config('use_jms_security')) %}
            $credentials = new Expression($credentials);
            {% endif %}
            return $this->checkCredentials('{{ column.credentials }}', ${{ builder.ModelClass }});
        {% else %}
            return true;
        {% endif %}
    }
```

without forgetting to add at top of the template:
```
{% if(admingenerator_config('use_jms_security')) %}
use JMS\SecurityExtraBundle\Security\Authorization\Expression\Expression;
{% endif %}twig
```

And the same things have to be done for filters [here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/FiltersType/type.php.twig#L60)

Does it make things clearer?

If you don't get it, don't worry, I see my mistake, I'll close your PR and make a commit to fix form generation this evening. And then, you could explain me what's your concrete situation, and see how to proceed (maybe there is a situation I didn't thought about?)

Let me know if something is still not clear enough,

tobias-93:
@sescandell I understand your reasoning with respect to the array/string thing.
However, I think I fixed the handling in the forms in a more elegant way (e.g. without changing all the form types but instead changing the `BaseType`).
If you think your way is better, I'm :+1: but I don't feel it is necessary.

sescandell:
Final result is the same.

It' just that taking into account the parameter on file generation, we avoid:
 - testing for each `checkCredentials` if parameter is true
 - "parameter injection" like [here](https://github.com/symfony2admingenerator/GeneratorBundle/pull/225/files#diff-023f16e06033fb912a5d7cb8461f2ef0R78)

tobias-93:
@sescandell Ok, I'll move the `Expression` generation to the subforms.

tobias-93:
@sescandell how about this?

Resources/templates/CommonAdmin/EditTemplate/fieldsets.php.twig
```diff
@@ -29,7 +29,7 @@
                             <div class="row">
                                 {% for column in rowColumns %}
                                     {% if column.credentials is not empty %}
-                                        {{ echo_if_granted(column.credentials, builder.ModelClass) }}
+                                        {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([column.credentials]) ~ ", " ~ builder.ModelClass ~ ')') }}
```
sescandell: Need to revert these changes: `is_granted` is enough here, `coumn.credentials` is a string (not an array)

Resources/templates/CommonAdmin/ListTemplate/FiltersBuilderTemplate.php.twig
```diff
@@ -50,7 +50,7 @@
                         {% for column in builder.filterColumns %}
                             {{ echo_block("list_filters_field_" ~ (column.name|lower)) }}
                             {% if column.filtersCredentials is not empty %}
-                                {{ echo_if_granted(column.filtersCredentials) }}
+                                {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([column.filterCredentials]) ~ ')') }}
```
sescandell: Same here

Resources/templates/CommonAdmin/ListTemplate/row.php.twig
```diff
@@ -11,7 +11,7 @@
 
         {% for column in builder.columns %}
             {% if column.credentials|default is not empty %}
-                {{ echo_if_granted(column.credentials, builder.ModelClass) }}
+                {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([column.credentials]) ~ ')') }}
```
sescandell: And here

Resources/templates/CommonAdmin/ListTemplate/thead.php.twig
```diff
@@ -32,7 +32,7 @@
 {% block list_thead_content_column %}
     {{ echo_block('list_thead_column_' ~ (column.name|lower)) -}}
     {% if column.credentials|default is not empty %}
-        {{ echo_if_granted(column.credentials) }}
+        {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([column.credentials]) ~ ')') }}
```
sescandell: Here too

Resources/templates/CommonAdmin/NestedListTemplate/thead.php.twig
```diff
@@ -32,7 +32,7 @@
 {% block list_thead_content_column %}
     {{ echo_block('list_thead_column_' ~ (column.name|lower)) -}}
     {% if column.credentials|default is not empty %}
-        {{ echo_if_granted(column.credentials) }}
+        {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([column.credentials]) ~ ')') }}
```
sescandell: Same here

Resources/templates/CommonAdmin/EditTemplate/fieldsets.php.twig
tobias-93: @sescandell the only problem then is that it will not be encapsulated in an `Expression`

Resources/templates/CommonAdmin/ShowTemplate/show.php.twig
```diff
@@ -28,7 +28,7 @@
                         {% for column in rowColumns %}
                             <div class="row">
                                 {% if column.credentials is not empty %}
-                                    {{ echo_if_granted(column.credentials, builder.modelClass) }}
+                                    {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([column.credentials]) ~ ')') }}
```
sescandell: Same here

Resources/templates/CommonAdmin/ShowTemplate/sidebar.php.twig
```diff
@@ -2,7 +2,7 @@
     {{ echo_block("show_sidebar") }}
         {% for name, widget in sidebar %}
             {% if widget.credentials|default is not empty %}
-                {{ echo_if_granted(widget.credentials, builder.ModelClass) }}
+                {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([widget.credentials]) ~ ')') }}
```
sescandell: Same here

Resources/templates/CommonAdmin/batch_actions.php.twig
```diff
@@ -13,7 +13,7 @@
                         {% for action in builder.BatchActions %}
                             {{ echo_block('batch_action_' ~ action.twigName) }}
                             {% if action.credentials %}
-                                {{ echo_if_granted(action.credentials) }}
+                                {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([action.credentials]) ~ ')') }}
```
sescandell: Same here

Resources/templates/CommonAdmin/generic_actions.php.twig
```diff
@@ -12,7 +12,7 @@
                 {% for action in builder.Actions %}
                     {{ echo_block("generic_action_" ~ action.twigName) }}
                     {% if action.credentials %}
-                        {{ echo_if_granted(action.credentials) }}
+                        {{ echo_if('is_one_admingenerator_granted(' ~ echo_twig_arr([action.credentials]) ~')') }}
```
sescandell: Same here

Resources/templates/CommonAdmin/object_actions.php.twig
```diff
@@ -12,7 +12,7 @@
         {% for action in builder.objectActions %}
             {{ echo_block('object_action_' ~ action.twigName) }}
             {% if action.credentials %}
-                {{ echo_if_granted(action.credentials, builder.ModelClass) }}
+                {{ echo_if("is_one_admingenerator_granted(" ~ echo_twig_arr([action.credentials]) ~ ", " ~ builder.ModelClass ~ ')') }}
```
sescandell: And here :)

Resources/templates/CommonAdmin/EditTemplate/fieldsets.php.twig
sescandell: Yes it is. In the Twig Extension: https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Twig/Extension/EchoExtension.php#L130

tobias-93: check, I'll revert :)

tobias-93:
@sescandell 
I've rebased the commits so I won't pollute your history, are you ok with the PR now?

sescandell:
Looks good to me, many thanks !!!

PR #227, opened 2016-01-08
Fix an error with AdmingenAllowed in form checking

tobias-93:
I'm sorry for bothering again, but I've made an error in the credentials checking in forms, leading to the `JMSSecurityBundle` wanting to parse the expression `'AdmingenAllowed'` which fails.
This is properly filtered out in this PR in my opinion.


sescandell:
:+1: thanks

PR #228, opened 2016-01-11
Fixes error in path generation and namespace generation when using pr…

tobias-93:
…efixes

### DIFF
--- a/Builder/Admin/EditBuilder.php
+++ b/Builder/Admin/EditBuilder.php
@@ -31,7 +31,7 @@ class EditBuilder extends BaseBuilder
             '%s\\%s\\Form\Type\\%s\\EditType',
             $this->getVariable('namespace_prefix'),
             $this->getVariable('bundle_name'),
-            $this->getModelClass()
+            $this->getBaseGeneratorName()
         );
     }
 }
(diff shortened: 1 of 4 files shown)


sescandell:
@tobias-93 

Could you please show me how do you "prefix" your generators?

Many thanks,

tobias-93:
@sescandell I use the built-in yaml-prefix functionality of the `GenerateAdminCommand`. For instance, my model is `Idb\EventBundle\Entity\Event` and in our case the generator is placed in `Idb\AdminGeneratorBundle\Resources\config\AdminEvent-generator.yml` (using `AdminEvent` as a prefix). Then, the FormTypes are placed in the `Idb\AdminGeneratorBundle\Form\Type\AdminEvent` namespace. Therefore, the prefix is not equal to an empty string or to the entity-name, but it includes an `Admin` prefix in our case.

sescandell:
... I'm used to use this format too... And everything works but probably because I'm usually using as a prefix my Entity... I'm just wondering so: are you sure about the `$model` concatenation after the prefix?

Don't we have to write: `$formsBundleNamespace = $generatorParameters['namespace_prefix'] . '\\' . $generatorParameters['bundle_name'] . '\\Form\\Type\\' . $generatorParameters['prefix']` (without the `$model` at the end according to the [generation process](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Generator/BundleGenerator.php#L162)

tobias-93:
@sescandell Yeah, my mistake... I'll update in a minute

sescandell:
Many thanks !

PR #246, opened 2016-02-03
Fix credentials checking in Excel builder when empty

tobias-93:
Not all credentials were checked for `AdmingenAllowed`, so it fails when using `JMSSecurityBundle`.
This PR fixes that.


bobvandevijver:
Many thanks!

PR #248, opened 2016-02-03
Embed Types

### LINKED ISSUE #247: EmbedType
I missunderstood requirements for embedTypes and didn't take enough care to previous implementation of this functionality. But PR #224 break this functionnality.

Before #224, we simply needed to create a generator file with only the "global" configuration.
Now, you need to create a full generator file with at least Edit and New builders. Doing that, corresponding controllers will be created too (that wasn't the case before the PR).

That should be fixed.

sescandell:
FIX #247

 - [x] Generate embed types from generators
 - [x] Auto-inject embed types

### DIFF
--- a/Generator/DoctrineODMGenerator.php
+++ b/Generator/DoctrineODMGenerator.php
@@ -47,6 +47,8 @@ class DoctrineODMGenerator extends Generator
         );
         $generator->setBaseGeneratorName($this->getBaseGeneratorName());
 
+        $this->generateEmbedTypes($generator);
+
         $builders = $generator->getFromYaml('builders', array());
 
         if (array_key_exists('list', $builders)) {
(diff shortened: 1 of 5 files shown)


Generator/DoctrineODMGenerator.php
```diff
@@ -47,6 +47,8 @@ protected function doBuild()
         );
         $generator->setBaseGeneratorName($this->getBaseGeneratorName());
 
+        $this->generateEmbedTypes($generator);
```
bobvandevijver: I miss the `typeBuilderNamespace` here

Generator/PropelGenerator.php
```diff
@@ -49,6 +49,8 @@ protected function doBuild()
         $generator->setColumnClass('Admingenerator\GeneratorBundle\Generator\PropelColumn');
         $generator->setBaseGeneratorName($this->getBaseGeneratorName());
 
+        $this->generateEmbedTypes($generator);
```
bobvandevijver: And here

sescandell:
@bobvandevijver yes... commit was pending :)

bobvandevijver:
Well, I was to fast. That's new :D

sescandell:
Well... I probably was tired... it's already working :)

PR cancelled

PR #252, opened 2016-02-06
Make routing update optional

### LINKED ISSUE #233: Optional routing update when using generate command
In some cases the automatic routing update is not wanted (for example, you want the files in their bundle, and not in the top routing). Before, it was possible to disable that, this is however no longer the case.

sescandell:
FIX #233

/cc @bobvandevijver


bobvandevijver:
Can't we directly add the functionality to specify the actual location of the routing file, as suggested by @pjam in #233?

sescandell:
Got it, I'll update the PR

tobias-93:
@sescandell I think you left a var_dump in the file in error. Maybe you should remove this?

sescandell:
OOps... many thanks!

bobvandevijver:
Looking good :+1:

PR #253, opened 2016-02-06
Add option --forms-only to generate-admin command

sescandell:
This PR propose to add the `--forms-only` option to the `generate-admin` command.

If set, only forms will be generated, and not controllers and views.
This is a kind of "improve process" for embed types (you had to manually remove all unnecessary generated files by hand)

### DIFF
--- a/Generator/DoctrineGenerator.php
+++ b/Generator/DoctrineGenerator.php
@@ -49,41 +49,48 @@ class DoctrineGenerator extends Generator
         );
         $generator->setBaseGeneratorName($this->getBaseGeneratorName());
 
-        $builders = $generator->getFromYaml('builders', array());
+        $builders = $generator->getBuildersFromYaml();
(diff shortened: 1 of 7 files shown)


Generator/DoctrineGenerator.php
```diff
@@ -49,41 +49,48 @@ protected function doBuild()
...
+        $formsOnly = $generator->isFormsOnly();
 
         if (array_key_exists('list', $builders)) {
-            $generator->addBuilder(new ListBuilderAction());
-            $generator->addBuilder(new ListBuilderTemplate());
+            if (!$formsOnly) {
+                $generator->addBuilder(new ListBuilderAction());
+                $generator->addBuilder(new ListBuilderTemplate());
+            }
             $generator->addBuilder(new FiltersBuilderType());
```
loostro: @sescandell why generate filter forms?

PR #255, opened 2016-02-10
Save and show action

bobvandevijver:
I've found some remnants in the code about the `save-and-show` action, so I've reinstated it.


sescandell:
:+1:

Issue #183, opened 2015-07-26
object actions template ok?


calvera:
hi,

i'm wondering if object actions template is 100% correct...

see https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ActionsAction/object_action.php.twig#L40

the return is never called...

also if I return from my objectActionXXX response (for example render page with form), it is not propagated and the page is not rendered.

I would propose an enhacement:

```
{% block attemptObjectAction %}

    /**
     * This function handles common object actions behaviour like
     * checking CSRF protection token or credentials.
     *
     * To customize your action look into:
     * executeObject{{ action.name|php_name|capitalize }}() - holds action logic
     * successObject{{ action.name|php_name|capitalize }}() - called if action was successfull
     * errorObject{{ action.name|php_name|capitalize }}()   - called if action errored
     */
    protected function attemptObject{{ action.name|php_name|capitalize }}($pk)
    {
        ${{ builder.ModelClass }} = $this->getObject($pk);

        try {
            {{ block('security_action_with_object') -}}

            {%- if action.csrfProtected %}
            if ('POST' == $this->get('request')->getMethod()) {
                {{ block('csrf_action_check_token') }}

                $response = $this->executeObject{{ action.name|php_name|capitalize }}(${{ builder.ModelClass }});

                if ($response) {
                    return $response;
                }

                return $this->successObject{{ action.name|php_name|capitalize }}(${{ builder.ModelClass }});
            }
            {% else %}
            $response = $this->executeObject{{ action.name|php_name|capitalize }}(${{ builder.ModelClass }});

            if ($response) {
                return $response;
            }

            return $this->successObject{{ action.name|php_name|capitalize }}(${{ builder.ModelClass }});
            {%- endif %}

        } catch (\Exception $e) {
            return $this->errorObject{{ action.name|php_name|capitalize }}($e, ${{ builder.ModelClass }});
        }
    }

{% endblock %}
```

what do you think?

-ks-

sescandell:
Hi @calvera 

The line you mentioned on your link is working only in the situation your objectAction is CSRF protected (meaning, require POST request) but the request made is a GET one. We so in that situation display a simple page with the object form action so the user can submit it with a POST request... (this is only if javacript fails, or user call the action in a new window or something like that).

Everything else should be enough to handle all your situations.
[...]

Issue #59, opened 2014-12-13
Provide AJAX capacities


sescandell:
We some times ago split templates to make AJAX possible.

On some client integration, I used this functionnality to add some AJAX capacities to the application. Usefull on list mainly (filters, object actions, pagination).
Let's see if it is possible to make it generic enough to be integrated into the bundle

PR #263, opened 2016-03-03
Use bootstrap modals instead of javascript confirm

calvera:
I'm not sure if you are interested in... 

this PR does not introduce BC break... nothing needs to be changed in the code...

please comment...


sescandell:
Hi @calvera 

Sounds like a good idea!

I'll check the PR ASAP ;)

Thanks,

Resources/templates/CommonAdmin/NestedListTemplate/NestedListBuilderTemplate.php.twig
```diff
@@ -21,3 +21,23 @@
...
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-default" data-dismiss="modal">{{ echo_trans('action.custom.cancel') }}</button>
+          <button type="submit" class="btn btn-primary">{{ echo_trans('action.custom.confirm') }}</button>
+      </div>
+      </form>
+    </div>
+  </div>
+</div>
+{{ echo_endblock() }}
```
bobvandevijver: This looks like a duplication of the code in the ListBuilderTemplate. So, it would be better to create a separate template which is included here: also makes it easier to customize :)

Resources/views/base_uncompressed.html.twig
```diff
@@ -85,6 +85,9 @@
         {# Wrapper end #}
         </div>
 
+    {% block modals %}
+    {% endblock modals %}
+
```
bobvandevijver: You've only added these in the uncompressed template? Why?

sescandell:
Hi,

I've seen some little things you need to change to make this working as expected.

`csrf-token` does not always exist. The action is not necessaraly protected. If you look at the previous code, we used a custom form only if the action was protected.

Previous behavior was:
 - If action is protected: prevent default event, generate a dynamic form and submit it
 - If action is not protected: let's default behavior continue

If I'm not wrong, in your new implementation:
 - If the action need a confirmation, you will always use a dynamic form: it might not be the case if the action is not protected
 - If the action doesn't need a confirmation but is protected, it doesn't use the dynamic form (xhereas it should use it)

Resources/templates/CommonAdmin/object_actions.php.twig
```diff
@@ -35,17 +35,23 @@
 
     {% spaceless %}
         {{ echo_spaceless() }}
+        <div {% if builder.yamlKey is same as('list') or builder.yamlKey is same as('nested_list') %}
```
sescandell: Are you sure this new "parent" doesn't break buttons display?
Given that object actions are wrapped by this code:

```twig
<div class="col-md-12 btn-group">
    {{ block('object_actions') }}
</div>
```

I'm wondering if:

```html
<div class="col-md-12 btn-group">
    <div>
        <a class="btn btn-default">button1</a>
        <a class="btn btn-default">button2</a>
        <a class="btn btn-default">button3</a>
    <div>
</div>
```

Would be rendered the same as:

```html
<div class="col-md-12 btn-group">
    <a class="btn btn-default">button1</a>
    <a class="btn btn-default">button2</a>
    <a class="btn btn-default">button3</a>
</div>
```

(Just a question, I don't remember how BS handle `btn-group`)

calvera:
thank you for your comments, i'll improve the PR

Resources/views/base_uncompressed.html.twig
calvera: base.html.twig extends base_uncompressed.html.twig, so I think adding to uncompressed is OK...

Resources/templates/CommonAdmin/NestedListTemplate/NestedListBuilderTemplate.php.twig
calvera: Good point, I will deduplicate

Resources/views/base_uncompressed.html.twig
bobvandevijver: You're right, my mistake.

Resources/templates/CommonAdmin/object_actions.php.twig
calvera: addition div is wrong, span would be better, but i will remove addional tag at all, which is the best

about dynamic forms:
I think, the correct behavior would be:
1) confirmation is needed -> show bs modal and POST a form with or without csrf protection if confirmed
2) no confirmation and csrf protection -> POST a dynamic form with csrf
3) no confirmation and no protection -> do not change, href link is followed

Do you agree?

sescandell: :+1:

calvera:
It doesn't work for generic action. Need more time to finish...

calvera:
so, now all should work, object actions, batch actions and generic actions as well. I have tested it on demo https://github.com/symfony2admingenerator/symfony2-admingenerator-demo-edition

Resources/public/js/main.js
```diff
@@ -13,33 +13,31 @@
             },
             options
         );
-        $(this.options.containerSelector).on('click', this.options.buttonSelector, this.clickHandler.bind(this));
+        $(this.options.containerSelector)
+            .find(this.options.buttonSelector)
+            .on('click',this.clickHandler.bind(this));
```
sescandell: Could you please come back to the previous implementation here. The difference is small, but previous version permit to handle AJAX loaded content (while the `find` version doesn't AFAIK).

Resources/templates/CommonAdmin/ActionsAction/ActionsBuilderAction.php.twig
```diff
@@ -24,6 +24,11 @@ class ActionsController extends BaseController
      */
     public function objectAction(Request $request, $pk, $action)
     {
+        // this is needed for confirmed object actions
+        if ($action == 'edit' || $action == 'show') {
+            return $this->redirectToRoute('{{ builder.baseActionsRoute }}_'.$action, array('pk' => $pk));
+        }
+
```
sescandell: Is this really  needed?

In which situation?

sescandell:
Hi @calvera 

I see you added the `confirmModal` property on `Action` objects, but I don't see `Action` class(es). Could you please double check that point (or remove the confirmModal option, that's your choice, I see what you mean by this parameter).

If you let the `confirmModal` option, please, update the documentation to recise there is a new option `confirmModal` in generators for actions, possible values and behavior depending on value.

Thank you,

Resources/templates/CommonAdmin/ActionsAction/ActionsBuilderAction.php.twig
calvera: i think yes, for example: on show page you have edit button which needs confirmation, object actions do POST...

Resources/public/js/main.js
calvera: yes, I will do, thanks for noticing this

Resources/templates/CommonAdmin/ActionsAction/ActionsBuilderAction.php.twig
sescandell: I think this is not necessary as edit / show actions are not mapped to object actions. The right routes should match first.

To be sure, you can also update your `show` modal handlers to dynamically update the $(form).prop('method') to get or post depending on protected or not action. That would be better.

calvera: if you make edit or show confirmed and protected it does not work without this...

it is bug in current code as well... try it in the demo...

calvera:
some rework, for demo see https://github.com/calvera/symfony2-admingenerator-demo-edition/tree/modals

sescandell:
Hi @calvera 

Everything seems good... I'm just wondering if your proposition about modals with additional fields is the right way to recommend. I agree with you: it works, and is fairly simple. But, with that solution I think you miss something important which is Symfony Form bidning and all validation process. What if the field required a valid unique email but is not valid (because already in the database). With this solution, the user doesn't have an easy way to manage that situation.

I already pushed a solution to that following an issue if I'm not wrong, based on Ajax, real form and managing validation constraints (actually based on Get / Post requests and form submitted or not). Could you please so complete your documentation based on that?

loostro:
I agree with @sescandell. The idea is good, but this should be improved to handle the form via AJAX and display server-side validation messages.

calvera:
@sescandell i'm not sure what you mean by 'I already pushed a solution to that following an issue...'. What you pushed and where? I have an example of full Symfony form in my symfony2-admingenerator-demo-edition/src/Admingenerator/DoctrineOrmDemoBundle/Resources/views/PostActions/index.html.twig

but this is not ideal... i'm not sure if we could somehow make all action twig templates ajax compatible...

sescandell:
Hi @calvera 

We discussed about that there is some times now [here](https://github.com/symfony2admingenerator/GeneratorBundle/issues/183#issuecomment-125242842) and following comment.

I took a look to your fork of the demo. In that code, you create a new modal with a form not binded to any Symfony2 Form ([here](https://github.com/calvera/symfony2-admingenerator-demo-edition/commit/eba1fa9132e56d029795a2fc7ba2b8299d6aa48c#diff-f8f8a54b447e59a0144a4b3f8b13b25cR5). My proposition is to manage that more globally with dynamic forms (as discussed [here](https://github.com/symfony2admingenerator/GeneratorBundle/issues/59#issuecomment-66908578) a long time ago).

Actually, if I clearly understand your fork, this is really near to how you handle `simpleedit action` (but loaders and form handlers are maybe missing).

If you prefer, I can propose you to split this PR in 2 parts: first one is moving to bootstrap modals (what you already made and is pretty good IMO) ; then, secondly, AJAX and dynamic form loading in modals (with documentation and/or helpers).

calvera:
@sescandell - the ajax/form part is now in cookbook only. I think someone could use the 'no-form' approach for simple task. it is his/her decision. let's make it clear in cookbook.

but I agree we could leave the full ajax/form for another PR.

thanks for comments and proposals...

p.s. after i read that discussion about ajaxification, i'll think about it.

calvera:
Where are we now? Am I supposed to do something more? :)

sescandell:
Looks good to me.

Ping @bobvandevijver and @loostro. If everybody is ok, I'll merge it :+1: 

Many thanks @calvera

sescandell:
@calvera could you please rebase the PR (otherwise, I'll do it :))

thanks

calvera:
squashed and rebased

sescandell:
Merged, many thanks @calvera :+1:

Issue #245, opened 2016-02-02
KnpMenuBundle as optional/recommended dependency


bobvandevijver:
Currently, the KnpMenuBundle is an required dependency and even when you do not use it, you need to install and enable it due to the [service definition](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/config/services.xml#L85). 

However, you can use the bundle fine without this bundle. Therefor, I would like to propose to make it an optional dependency and to create the service definition such that it can be enabled if wanted. Any thoughts?

sescandell:
You would also have to [update the template here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/views/Sidebar/layout.html.twig#L14) if you do so.

Is there any situation you would don't want to use it?

bobvandevijver:
I do not use most of the default templates, but I imagine there will be more occurrences, so that must be checked.

We do not use the menu bundle due to the fact we integrated the bundle in a much larger application, which does not use the KnpMenuBundle for its menu. Actually, we use a self-written menu generator and we define the menu completely in the routing files. But, we might be alone is this :)

sescandell:
Well... maybe we could remove it from the bundle you're right, and just precise in a "cookbook" how to easily integrate KnpMenu in the bundle...

@loostro any insight?

loostro:
I'm :+1: for removeing the dependency. A dashboard/links system is not the "core" of this bundle. The core of this bundle is generateing CRUD for models, and auto-generateing routes is what I still consider "core", but the menu system is purely optional. It should be described how to quickly setup basic menu system with KnpMenuBundle, but it's not mandatory to use it.

bobvandevijver:
There already is some doc in the cookbook about the menu, but it is probably a little bit outdated (see #239). I will copy the current doc to the new documentation, but I will not adjust it yet as this should be inspected first.

PR #265, opened 2016-03-06
Removed iCheck dependency

bobvandevijver:
For as far as I can see, there was only one place where the iCheck was used, which was also in fault. 

This PR removes the complete iCheck dependency, as I believe it is not necessary. @sescandell, you were the first to add the iCheck to the `bower.json`, what's your opinion?

### DIFF
--- a/Resources/public/js/main.js
+++ b/Resources/public/js/main.js
@@ -76,7 +76,7 @@
         $(this.options.containerSelector).find('*[type=submit]').hide();
 
         this.actionInputSelector.on('change', this.selectedActionChangedHandler.bind(this));
-        this.allElementsToggleButton.on('ifChecked ifUnchecked', this.allElementsChangedHandler.bind(this));
+        this.allElementsToggleButton.on('change ifChecked ifUnchecked', this.allElementsChangedHandler.bind(this));
     };
 
     S2A.batchActionsManager.prototype = {
@@ -104,9 +104,11 @@
         },
 
         allElementsChangedHandler: function(evt){
-            $(this.options.elementSelector)
-                .prop('checked', $(evt.currentTarget).is(':checked'))
-                .iCheck('update');
+            var $element = $(this.options.elementSelector);
+            $element.prop('checked', $(evt.currentTarget).is(':checked'));
+            if(typeof($element.iCheck) == "function"){
+                $element.iCheck('update');
+            }
         },
 
         isValidActionSelected: function(actionValue){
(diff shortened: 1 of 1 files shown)


sescandell:
Actually, iCheck is a dependency of AdminLTE. That is thanks to this plugin that checkboxes are rendered with a better look in all templates. Maybe there is some places it is mal-initialized and not working properly?

bobvandevijver:
@sescandell I do not use the AdminLTE templates, but when I search this repository I do not find any iCheck call to initialize the checkboxes. 

In the AdminLTE changelog it states that the iCheck dependency is removed? [AdminLTE changelog](https://github.com/almasaeed2010/AdminLTE/blob/4cd707a49e7a8d9a7a740de9daffc23018844317/changelog#L90). Also I can not find any `.iCheck()` call in the AdminLTE repo... But they do mention in their documentation that they use it. 

I will change this and check for the existence of the function before calling, that should provide enough compatibilty :)

bobvandevijver:
@sescandell How about this?

sescandell:
Actually, they do use it:

 - On the login page for example: https://almsaeedstudio.com/themes/AdminLTE/pages/examples/login.html
 - Dependency is in the plugin directory in their repo: https://github.com/almasaeed2010/AdminLTE/tree/master/plugins/iCheck
 - And initialization is automatically made on all checkboxes by default.

I'm not sure `change` event is working... did you had any chance to test it? Otherwise, I'll do on the demo application.

Thanks,

bobvandevijver:
I did not check the plugins directory, just the `bower.json`: in my opinion copied dependencies are not the way to go... But okay, I see it's used. 

The change event is the default event in jQuery, so I've no reason to suspect it's not working. It is working in my environment, but I did not test it on demo.

Looks like there was a problem with the `change` event not firing ([icheck#244](https://github.com/fronteed/icheck/issues/244)), but it should be solved ([icheck#259](https://github.com/fronteed/icheck/pull/259).

sescandell:
Great so :+1:

bobvandevijver:
@sescandell Did you test it?

sescandell:
Not yet,

sescandell:
Going to test it now,

sescandell:
Well, actually, I was wrong: iCheck is not enabled by default. You need to explicitly initialize it. So your fix is good, if someone wants to use `iCheck`, he must active it by itself (given that it is already there...).

I agree with you: adminlte doesn't manage frontend dependencies well...

sescandell:
Well, actually I just check the previous version and... it's already not working. The reason is the dis version of Adminlte doesn't include the icheck initialization code (found in `adminlte/js/app.js` but not in `admin-lte/dist/js/app.js`).

Missing code:

```js
    /*
     * We are gonna initialize all checkbox and radio inputs to
     * ckck plugin in.
     * You can find the documentation at http://fronteed.com/iCheck/
     */
    $("input[type='checkbox']:not(.simple), input[type='radio']:not(.simple)").iCheck({
        checkboxClass: 'icheckbox_minimal',
        radioClass: 'iradio_minimal'
    });
```

I totally agree with you: AdminLTE doesn't handle frontend dependencies the right way.

sescandell:
All of that to say: you cant merge your PR, functionnality was already broken, I'll test to integrate it in a second time.

sescandell:
That said, I tried your fix with the iCheck plugin correctly initialized, and it breaks the toggle selector on lists.

loostro:
@sescandell maybe we should open an issue (or PR) in AdminLTE project  about incorrect handleing of dependencies?

bobvandevijver:
Guess it's to much work for us to fix the complete AdminLTE dependency handling ;)

@sescandell Do you have any info on why it breaks when the iCheck is initialized?

sescandell:
> @sescandell Do you have any info on why it breaks when the iCheck is initialized?

If I'm not wrong, this is because the `change` event is not triggered when iCheck is enabled on the concerned checkbox. That's why `ifcheck` and `ifuncheck` where binded. Maybe with `change ifcheck ifuncheck` it works... I'll try that when I have 2 minutes.


 > @sescandell maybe we should open an issue (or PR) in AdminLTE project about incorrect handleing of dependencies?

Someone already tried that, but PR was refused: https://github.com/almasaeed2010/AdminLTE/pull/548

bobvandevijver:
> If I'm not wrong, this is because the change event is not triggered when iCheck is enabled on the concerned checkbox. That's why ifcheck and ifuncheck where binded. Maybe with change ifcheck ifuncheck it works... I'll try that when I have 2 minutes.

Well, than the comment I made earlier probably did not yet make it into adminLTE... 
> Looks like there was a problem with the change event not firing (icheck#244), but it should be solved (icheck#259).

Guess adding the check to the event list should solve it, but than it might also be the case that when the iCheck plugin is updated the event is fired twice. However, with the current implementation of the `allElementsChangedHandler` that should not really matter.

bobvandevijver:
@sescandell Did you have any chance to test it?

sescandell:
Checking it now...

sescandell:
@bobvandevijver I just tested various combinations. Finally, if we want to be coherent with how the bundle should work initially, here is the right `main.js` file (IMO):

```js
    ...
    this.allElementsToggleButton.on('change ifChecked ifUnchecked', this.allElementsChangedHandler.bind(this));
    ...

// Called once the DOM is loaded
$(function(){
    // Initializing the iCheck dependency: should be made in AdminLTE dist app.js code...
    $("input[type='checkbox']:not(.simple), input[type='radio']:not(.simple)").iCheck({
        checkboxClass: 'icheckbox_minimal',
        radioClass: 'iradio_minimal'
    });
    $("input[type='checkbox']:not(.simple), input[type='radio']:not(.simple)").on('ifChecked ifUnchecked', function(){
            $(this).trigger('change');
        });
});
```

That way:

 - iCheck is correctly initialized
 - users can simply bind on `change` event to do what they want


Is that good for you?

bobvandevijver:
@sescandell I'm not sure about this, but I believe we somewhere earlier said that this project should be usable without the AdminLTE dependency as it is not necessary (I do not use it for instance :smile:).

So, I took your sample and adjusted it such that it checks if the iCheck functionality is available. I've added it and squashed the PR. 

This solution means that it has another check, but in general that should not hurt anyone (actually it helps me, as otherwise the script fails). Furthermore, I've moved more of the initialization calls into the `$(function(){})`, such that they are also loaded when they need to be (no longer the need to put it in the bottom of the DOM). 

I believe in this way we help every user, so, what do you think?

sescandell:
> I believe we somewhere earlier said that this project should be usable without the AdminLTE dependency as it is not necessary (I do not use it for instance :smile:).

I don't know where did you read that, but, AFAIK, that's not true. Removing the AdminLTE dependency would break many things: design, menu, navigation, and probably others. I agree we can use the bundle without the AdminLTE, but removing JS / CSS from AdminLTE would break many things.


That said, regarding the PR: :+1:

bobvandevijver:
Guess it's a design choice, but I found the response I was referring to (from @loostro in https://github.com/symfony2admingenerator/GeneratorBundle/issues/245#issuecomment-179923057 about the KnpMenuBundle)
> I'm :+1: for removeing the dependency. A dashboard/links system is not the "core" of this bundle. The core of this bundle is generateing CRUD for models, and auto-generateing routes is what I still consider "core", but the menu system is purely optional. It should be described how to quickly setup basic menu system with KnpMenuBundle, but it's not mandatory to use it.

So, I would say that the AdminLTE should be available if you want to use it, but we should not force it (and make it a dependency). I also doubt a lot of stuff would break, but it doesn't matter for now. 

I'm merging it and releasing 2.1, thanks for the feedback :smiley:

Issue #266, opened 2016-03-07
PagerfantaBundle update breaks admingenerator pagination


bobvandevijver:
In [PagerfantaBundle#131](https://github.com/whiteoctober/WhiteOctoberPagerfantaBundle/pull/131) an update is created which by default omits the `page` URL param if the page is 1. This is apparently also correct according to the RFC, although it breaks our pagination. 

As we save the current page in the session, it is no longer possible to return to the first page of the list after you've visited another page (when using the paginator links, you can change a page=4 url manually to 1). We save the page in the session such that we can easily return to the same page after an specific action. 

I see four possible solutions:
1. Fix the Pagerfanta version on the old version
2. Make the Pagerfanta behavior configurable in the Pagerfanta bundle self, which can be easily done with a single parameter.
3. Adapt our [AdminGeneratorView](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Pagerfanta/View/AdmingeneratorView.php#L113) to overwrite the Pagerfanta behavior for this first page
4. Adapt our complete session model and update all URL used everywhere.

Option 1 & 4 are a no go in my opinion. I will try to push option 2 to the Pagerfanta bundle, but if it fails we probably use option 3.

bobvandevijver:
I've created a PR at [PagerfantaBundle#140](https://github.com/whiteoctober/WhiteOctoberPagerfantaBundle/pull/140) to fix this, when merged we only need to add the param to our Pagerfanta call.

sescandell:
:+1: thanks for the prevention

bobvandevijver:
Well, it's merged. Now we just have to wait for a new release, so that I can update our composer file :)

Issue #226, opened 2016-01-07
Variables don't get injected in Twig Extensions when generating bundle


tobias-93:
Unfortunately, I still experience problems related to #225. When operating, the `use_jms_security` parameter is injected fine. However, when generating, the parameter doesn't get injected due to the code in https://github.com/symfony2admingenerator/TwigGenerator/blob/master/src/TwigGenerator/Builder/BaseBuilder.php#L344.

I see three possible options for fixing this in this case:

1. The original solution in #225, so replacing all `echo_if_granted` by `echo_if('is_one_granted'`;
2. Injecting the variable into the function as a variable, thereby changing the function signature;
3. Finding out how to inject the variables when generating the bundle, but I'm not sure how to do that.

@sescandell what do you think?

sescandell:
Hi @tobias-93 

Let me check that. I'll let you know quickly

sescandell:
Ok found.

There was something strange that bother me: how could all other `twig functions` works whereas they are not in [this initialization](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Builder/BaseBuilder.php#L35). There was no reason `EchoExtension` doesn't know the `use_jms_security` (why `ConfigExtension` is so working well?).

And the thing is, the mistake __is__ the initialization. We just need to remove it. Twig extensions are copied from the "global" `twig` service [here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Generator/DoctrineGenerator.php#L45), [here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Generator/DoctrineODMGenerator.php#L43) and [here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Generator/DoctrineODMGenerator.php#L43).

Tested on the demo project: working well :+1: Commit is coming (thanks for the report).

tobias-93:
@sescandell Great, thanks so much for looking into it!

PR #267, opened 2016-03-07
PagerFanta return to page no. 1 fix

laszlo-karpati:
Fix bug #266


bobvandevijver:
This does not fix #226, as every time your return from for example a delete action, you get back on page 1. However, you expect to be returned to the page you were when deleting the object.

laszlo-karpati:
Oh, sorry i missed that!

PR #270, opened 2016-03-09
Forward to list on missing object during deletion

bobvandevijver:
When a object is not found during deletion (for example, when clicking the confirm button twice on the confirm page), you get an Exception which is too general: Object could not be removed. 

This PR introduces an automatic forward, which creates a notification that it could not be removed as it already was removed. 

Also added some tweaks in parts of the code to make it more consistent.


tobias-93:
I'm :+1: for this, an Exception is too harsh in this case.

Resources/translations/Admingenerator.fr.yml
```diff
@@ -70,6 +70,7 @@ action.object.delete:
     confirm:  Etes-vous sur de vouloir supprimer cet élement?
     success:  L'élément sélectionné a été supprimé.
     error:    L'élément sélectionné ne peut pas être supprimé.
+    notfound: Element ne peut pas être supprimé car il est déjà supprimé.
```
sescandell: French one: *Suppression impossible : l'élément est inconnu*

Resources/templates/Doctrine/ActionsBuilderAction.php.twig
```diff
@@ -12,7 +12,7 @@
         ${{ builder.ModelClass }} = $this->getObjectQuery($pk)->getOneOrNullResult();
 
         if (!${{ builder.ModelClass }}) {
-            throw new \InvalidArgumentException("No {{ model }} found on {{ builder.getFieldGuesser().getModelPrimaryKeyName(model) }} : $pk");
+            throw new NotFoundHttpException("No {{ model }} found on {{ builder.getFieldGuesser().getModelPrimaryKeyName(model) }} : $pk");
```
sescandell: Take care, this is a BC break that need a comment in the UPDATE doc.

sescandell:
2 comments, otherwise : :+1:

Resources/templates/Doctrine/ActionsBuilderAction.php.twig
bobvandevijver: Not really, as by default it is catched. Only when you use the new `throw_exception` parameter the thrown exception changes. Nevertheless, I will add a note.

Resources/translations/Admingenerator.fr.yml
bobvandevijver: Done.

sescandell:
:+1: good for me

PR #272, opened 2016-03-09
Fixes #245. Removed KnpMenuBundle as dependency.

### LINKED ISSUE #245: KnpMenuBundle as optional/recommended dependency

bobvandevijver:
This PR removes the dependency of the KnpMenuBundle and add the documentation/upgrade notes on how to keep using it like earlier. 

This is a BC-breaking PR, but with the documentation nobody should have a problem. This means that I will merge this when we are ready to release 2.1.

@loostro @sescandell Please review :)


DependencyInjection/Configuration.php
```diff
@@ -255,6 +255,7 @@ public function getConfigTreeBuilder()
                 ->scalarNode('exit_route')->defaultNull()->end()
                 ->scalarNode('generator_cache')->defaultNull()->end()
                 ->scalarNode('default_action_after_save')->defaultValue('edit')->end()
+                ->scalarNode('knp_menu_alias')->defaultNull()->end()
```
sescandell: You need to update the Test class too

bobvandevijver: Added!

sescandell:
:+1:

PR #273, opened 2016-03-10
Fixes #271. Added server validation for 'None' action

### LINKED ISSUE #271: Should we catch empty batch actions?
Currently, when JS fails a button is displayed for the batch actions, which simply executes the selected action. 

However, when the 'Selected' is selected (so no action) and the button is clicked, the following error is thrown:
```
Undefined "attemptBatchNone" method. Does batch action "none" exist in your generator file?
```

This indicates that there is no server check for faulty requests here. We should probably check this in the controller and forward the user to the list page when this happens. 

Any opinions? @sescandell @loostro?

bobvandevijver:
See #271 for details on why.

### DIFF
--- a/Resources/translations/Admingenerator.fr.yml
+++ b/Resources/translations/Admingenerator.fr.yml
@@ -80,7 +80,7 @@ action.batch.delete:
     success:  Les enregistrements ont été supprimés avec succès.
     error:    Les enregistrements n'ont pu être supprimés.
 
-action.batch.none: Premièrement, vous devez sélectionner une action.
+action.batch.none: Merci de sélectionner une action.
 
 # Custom actions
 action.custom:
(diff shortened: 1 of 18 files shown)


Resources/translations/Admingenerator.fr.yml
```diff
@@ -80,6 +80,8 @@ action.batch.delete:
     success:  Les enregistrements ont été supprimés avec succès.
     error:    Les enregistrements n'ont pu être supprimés.
 
+action.batch.none: Premièrement, vous devez sélectionner une action.
```
sescandell: French one would be: *"Merci de sélectionner une action"*

sescandell:
:+1:

Resources/translations/Admingenerator.fr.yml
bobvandevijver: Done!

sescandell:
Thanks

PR #275, opened 2016-03-18
fixing dbType array when using propel

mirza99:
When using Propel, and having a column with the type "array", in the function getOptions the parameter type is CollectionType.
When having a CollectionType, it is considering that we have always a relation but in the case using an array we havent. So a exception is thrown saying that we are calling getType on a boolean variable.

The fix is :
```

if (preg_match("#CollectionType$#i", $type)) {
            $relation = $this->getRelation($columnName, $class);

++            if ($relation) {
                return array(
                    'allow_add'     => true,
                    'allow_delete'  => true,
                    'by_reference'  => false,
                    'entry_type' => 'entity',
                    'entry_options' => array(
                        'class' => \RelationMap::MANY_TO_ONE === $relation->getType() ? $relation->getForeignTable()->getClassname() : $relation->getLocalTable()->getClassname()
                    )
                );
++            }
            
++            return array(
++                   'entry_type' => 'text',
++               );
        }


```

### DIFF
--- a/Guesser/PropelORMFieldGuesser.php
+++ b/Guesser/PropelORMFieldGuesser.php
@@ -325,15 +325,21 @@ class PropelORMFieldGuesser
         if (preg_match("#CollectionType$#i", $type)) {
             $relation = $this->getRelation($columnName, $class);
 
+            if ($relation) {
+                return array(
+                    'allow_add'     => true,
+                    'allow_delete'  => true,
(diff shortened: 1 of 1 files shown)


sescandell:
Many thanks @mirza99 :+1:

mirza99:
You're welcome ;)

Issue #186, opened 2015-07-29
Issue with generatorBundle and FosUser : No mapping found for field 'roles'


rpostolov:
Hi everybody !

I have some problem using FOSUserBundle.

I am on symfony 2.7.1 and i installed fosuserbundle  "~2.0@dev"

I create bundle User and admin but when i try to go to edit admin or else i have this problem

```
No mapping found for field 'roles' on class 'Ip\UserBundle\Entity\User'.
```

Can someone help me cause i'm stuck on this bug a long time already.

I posted this issue on FosUser Repository but maybe it's an problem from GeneratorBundle.

I use FosUser on Entity User :

```
use FOS\UserBundle\Model\User as BaseUser;

use Doctrine\ORM\Mapping as ORM;

/**
 * User
 *
 * @ORM\Table(name="user")
 * @ORM\Entity(repositoryClass="Ip\UserBundle\Entity\UserRepository")
 * @ORM\InheritanceType("JOINED")
 * @ORM\DiscriminatorColumn(name="type", type="string")
 * @ORM\DiscriminatorMap({"admin" = "Ip\AdminBundle\Entity\Admin", "customer" = "Ip\CustomerBundle\Entity\Customer"})
 * 
 * @ORM\HasLifecycleCallbacks() 
 */
abstract class User extends BaseUser```

And Here is for Entity Admin : 

```
namespace Ip\AdminBundle\Entity;

use Doctrine\ORM\Mapping as ORM;
use PUGX\MultiUserBundle\Validator\Constraints\UniqueEntity;

use Ip\UserBundle\Entity\User as User;

/**
 * Admin
 *
 * @ORM\Table(name="admin")
 * @ORM\Entity(repositoryClass="Ip\AdminBundle\Entity\AdminRepository")
 * @UniqueEntity(fields = "username", targetClass = "Ip\UserBundle\Entity\User", message="fos_user.username.already_used")
 * @UniqueEntity(fields = "email", targetClass = "Ip\UserBundle\Entity\User", message="fos_user.email.already_used")
 */
class Admin extends User```

Here is my configuration in config.yml

```
# Doctrine Configuration
doctrine:
    dbal:
        driver:   pdo_mysql
        host:     "%database_host%"
        port:     "%database_port%"
        dbname:   "%database_name%"
        user:     "%database_user%"
        password: "%database_password%"
        charset:  UTF8
       
    orm:
        auto_generate_proxy_classes: "%kernel.debug%"
                
[...]

PR #284, opened 2016-05-23
Issue #186 : allows "array" db type to be used with a collection form type

pierrerolland:
Collection types currently have to have a Doctrine mapping association. That makes simple array type unable to be used as a collection type. 

In this "fix", we check, when there is a collection type, if the column has a mapping association before getting it, to avoid the "No mapping found" error reported here : https://github.com/symfony2admingenerator/GeneratorBundle/issues/186

### DIFF
--- a/Guesser/DoctrineFieldGuesser.php
+++ b/Guesser/DoctrineFieldGuesser.php
@@ -286,17 +286,21 @@ abstract class DoctrineFieldGuesser
         }
 
         if (preg_match("/CollectionType$/i", $type)) {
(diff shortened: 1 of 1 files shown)


sescandell:
Thanks @pierrerolland

pierrerolland:
You're welcome, @sescandell :)

Hasn't it been corrected for Propel guesser by this commit ? https://github.com/symfony2admingenerator/GeneratorBundle/commit/0d6c9a89fe53e79c4261bcf3e52bce512c477476

sescandell:
You're right!

Thanks again

PR #289, opened 2016-05-31
Fixed #288, auto generated form types now have unique service name

### LINKED ISSUE #288: Possible bug in Admingenerator\GeneratorBundle\DependencyInjection\AdmingeneratorGeneratorExtension::registerFormsServicesFromGenerator

Pattern is: 

- `\One\ModelX`
- `\Two\ModelX`

TheCelavi:
Fixed #288, auto generated form types now have unique service name, there is no possibility for overwriting service name one over the other.

### DIFF
--- a/DependencyInjection/AdmingeneratorGeneratorExtension.php
+++ b/DependencyInjection/AdmingeneratorGeneratorExtension.php
@@ -269,13 +269,15 @@ class AdmingeneratorGeneratorExtension extends Extension
     private function registerFormsServicesFromGenerator(array $generatorParameters, array $builders, ContainerBuilder $container)
     {
         $modelParts = explode('\\', $generatorParameters['model']);
-        $model = array_pop($modelParts);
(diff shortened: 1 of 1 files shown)


DependencyInjection/AdmingeneratorGeneratorExtension.php
```diff
@@ -269,13 +269,15 @@ private function registerGeneratedFormsAsServices(ContainerBuilder $container)
     private function registerFormsServicesFromGenerator(array $generatorParameters, array $builders, ContainerBuilder $container)
     {
         $modelParts = explode('\\', $generatorParameters['model']);
-        $model = array_pop($modelParts);
+        $model = strtolower(array_pop($modelParts)); // @TODO: BC Support, remove it starting from v.3.0.0
```
TheCelavi: Minor optimisation, no need to call 3 times "strtolower". Comment is added for future release when BC is ok.

TheCelavi:
Wait a sec, I can not believe that I forgot something...

PR #291, opened 2016-05-31
Fixed #288, auto generated form types now have unique service name, no possibility for overwritting.

### LINKED ISSUE #288: Possible bug in Admingenerator\GeneratorBundle\DependencyInjection\AdmingeneratorGeneratorExtension::registerFormsServicesFromGenerator

loostro:
Squashed #290

### DIFF
--- a/DependencyInjection/AdmingeneratorGeneratorExtension.php
+++ b/DependencyInjection/AdmingeneratorGeneratorExtension.php
@@ -269,13 +269,15 @@ class AdmingeneratorGeneratorExtension extends Extension
     private function registerFormsServicesFromGenerator(array $generatorParameters, array $builders, ContainerBuilder $container)
     {
         $modelParts = explode('\\', $generatorParameters['model']);
-        $model = array_pop($modelParts);
+        $model = strtolower(array_pop($modelParts)); // @TODO: BC Support, remove it starting from v.3.0.0
+        $fullQualifiedNormalizedModelName = strtolower(str_replace('\\', '_', ltrim($generatorParameters['model'])));
         $formsBundleNamespace = sprintf(
             '%s%s\\Form\\Type\\%s',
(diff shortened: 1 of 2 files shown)


loostro:
@sescandell Was the service definition feature documented anywhere? I couldnt find it. If it was, we should also update the docs.

TheCelavi:
@loostro This one is missing code, not correct PR. Oh man, I made a mess... #290 has proper code in it.

loostro:
@TheCelavi I copied the missing upgrade notes. I don't see anything else missing from your PR.

TheCelavi:
@loostro 

Line 273:

    $fullQualifiedNormalizedModelName = strtolower(str_replace('\\', '_', ltrim($generatorParameters['model'])));

Function `ltrim` is missing second parameter which should be double backslash (\\)

    $fullQualifiedNormalizedModelName = strtolower(str_replace('\\', '_', ltrim($generatorParameters['model'], '\\')));

sescandell:
> @sescandell Was the service definition feature documented anywhere? I couldnt find it. If it was, we should also update the docs.

Do you mean explanation on how all of this is working? or "generated names"?

TheCelavi:
@loostro In #292 I have fixed the mess which I caused yesterday, sorry. I have just committed a cookbook entry as well in regards dependency injection for form types.

@sescandell In cookbook entry I have tried to explain in details how service names are generated for form types, so take a look at [https://github.com/RunOpenCode/GeneratorBundle/blob/82d0cec04bc723d26093097d66ba1070fac00875/Resources/doc/cookbook/inject-services-into-form-types.md](https://github.com/RunOpenCode/GeneratorBundle/blob/82d0cec04bc723d26093097d66ba1070fac00875/Resources/doc/cookbook/inject-services-into-form-types.md) and see if this is documented good enough.

loostro:
Closed in favor of #292

PR #292, opened 2016-06-01
Fixed #288, auto generated form types now have unique service name.

### LINKED ISSUE #288: Possible bug in Admingenerator\GeneratorBundle\DependencyInjection\AdmingeneratorGeneratorExtension::registerFormsServicesFromGenerator

TheCelavi:
@loostro Sorry for mess up from yesterday, I had a lot of thing on my plate so I was making those stupid mistakes and git --amend does not work as I thought it does...

So here is new PR with one commit point. You can close previous one.

Sorry again.


sescandell:
Looks good for me 

/cc @loostro @bobvandevijver

bobvandevijver:
:+1:

loostro:
@TheCelavi great job, thanks :)

PR #293, opened 2016-06-05
Finder short circuit if path is not a dir

loostro:
I think we should check here for `is_dir` instead of simply `file_exists`. If there is no directory, there is no need to instantiate Finder object.


sescandell:
:+1: makes sense

bobvandevijver:
:+1: Sure, why not. Although when you have a `config.yml` and a `services.yml` it still does not help :sweat_smile:

PR #294, opened 2016-07-13
Add information to errors during warmup, to ease debugging

tobias-93:
Also removes some checks for exception types which aren't handled differently.


bobvandevijver:
Merged, thanks!

Issue #296, opened 2016-07-22
[RFC] Implement Symfony3.2 workflow component


loostro:
It seems like symfony 3.2 workflow component could be a perfect addition to admingenerator. We should think how it could be easily implemented into out tempaltes without breaking current behaviour. Or maybe it's worth a BC break? Please share your thoughts.

javiereguiluz:
I'd like to ask you in which kind of operations/tasks are you thinking about integrating the new Workflow component. Thanks!

bobvandevijver:
For reference: the [Symfony 3.2 Workflow component](http://symfony.com/blog/new-in-symfony-3-2-workflow-component)

I'm also not sure where we would want to use it. You want to add the workflow to the editing of the bundle? I'm not sure if that's something we would want, imho it is out of the scope of this bundle.

loostro:
I meant adding support `workflow_can` checks instead/in addition to(?) of security `is_granted` checks.

bobvandevijver:
That's is not where the workflow component is meant to be used for if I understand it correctly. How does a state change if you want to view the list? Or if you simply only edit the object? 

I would stay with the `is_granted`-checks.

loostro:
True :)

PR #297, opened 2016-07-23
Implement symfony 3.2 workflows

ioleo:
- [x] add twig function for `workflow_can`
- [x] add new action parameter `workflows` (defaults to empty array)
- [x] wrap object actions in `if workflow_can` blocks if workflows are defined
- [x] update docs

Related to #296


loostro:
This is a draft, not yet tested.

PR #298, opened 2016-07-23
Fixed field guessers when it comes to manyToMany associations.

TheCelavi:
Fixed column guessers, comments in code explains errors and fixes.

### DIFF
--- a/Guesser/DoctrineFieldGuesser.php
+++ b/Guesser/DoctrineFieldGuesser.php
@@ -62,7 +62,8 @@ abstract class DoctrineFieldGuesser
     public function getManyToMany($model, $fieldPath)
     {
         $resolved = $this->resolveRelatedField($model, $fieldPath);
-        return !$this->getMetadatas($resolved['class'])->isAssociationWithSingleJoinColumn($resolved['field']);
+        $metadata = $this->getMetadatas($resolved['class']);
+        return $metadata->hasAssociation($resolved['field']) && !$metadata->isAssociationWithSingleJoinColumn($resolved['field']);
     }
 
     /**
--- a/Guesser/PropelORMFieldGuesser.php
+++ b/Guesser/PropelORMFieldGuesser.php
@@ -73,6 +73,8 @@ class PropelORMFieldGuesser
         if ($relation) {
             return \RelationMap::MANY_TO_MANY === $relation->getType();
         }
+
+        return false;
     }
 
     /**


Guesser/PropelORMFieldGuesser.php
```diff
@@ -73,6 +73,8 @@ public function getManyToMany($model, $fieldPath)
         if ($relation) {
             return \RelationMap::MANY_TO_MANY === $relation->getType();
         }
+
+        return false;
```
TheCelavi: Method return statements are inconsistent, on line 74 it will return "TRUE", otherwise, method is type of "void". 

Method is declared to return boolean, so this makes method to be consistent with its signature.

Guesser/DoctrineFieldGuesser.php
```diff
@@ -62,7 +62,8 @@ public function getAllFields($class)
     public function getManyToMany($model, $fieldPath)
     {
         $resolved = $this->resolveRelatedField($model, $fieldPath);
-        return !$this->getMetadatas($resolved['class'])->isAssociationWithSingleJoinColumn($resolved['field']);
+        $metadata = $this->getMetadatas($resolved['class']);
+        return $metadata->hasAssociation($resolved['field']) && !$metadata->isAssociationWithSingleJoinColumn($resolved['field']);
```
TheCelavi: We have to check if if association even exists, before we check if that is many-to-many association. 

Otherwise, this method would return "true" if association does not exists, because, if there is no association, it is not association with single join column -> which will get result "it is many-to-many" -> which is wrong.

bobvandevijver: This seams like a BC-breaking behavior change. The only occurrence of usage seems to be [here](https://github.com/symfony2admingenerator/GeneratorBundle/blob/master/Resources/templates/CommonAdmin/ListAction/filters.php.twig#L88). Can you tell me if you tested this with the filters? I'm currently have no possibility to test...

TheCelavi: It is not BC break - it is a bugfix. Currently, a wrong code is generated, that is why I have submitted this patch. 

Per example, for all non-relation columns you would get this code:

    protected function filterColumnName($queryFilter, $value)
    {
        $queryFilter->addXXXFilter('columnName', $value, 1);
    }

and if you have date/datetime field -> you are in trouble, because "addDateFilter" as third parameter has date format as string.

Probably no one noticed a bug so far, but it is a bug.

bobvandevijver: Bug fixes can also be BC-breaking :wink: But I see the problem now, thanks for the fix!

TheCelavi: Sad true :D

PR #299, opened 2016-07-26
Generic actions EDIT, SHOW do have access to object, and security can…

TheCelavi:
… depend on accessibility of that object. This was a bug. Now it is fixed.

Let me explain:

Object actions do depend on object, so you can check ACL (or custom voter) if you are going to allow user to do some action (Edit, Delete, Show).

Generic actions however do allow us to check for security (per example ROLES) but object is not passed to context. This is especially a drawback when using JMS Security Extra.

Reason for that is, of course, a "New" action, object is not available in context.

But -> solution is simple, in template we can use ternary operator, and pass a null when object is not available in context. 

BTW -> git is being a bitch for this file, so I am going to mark lines of code which are changed.

### DIFF
--- a/Guesser/DoctrineFieldGuesser.php
+++ b/Guesser/DoctrineFieldGuesser.php
@@ -62,7 +62,8 @@ abstract class DoctrineFieldGuesser
     public function getManyToMany($model, $fieldPath)
     {
         $resolved = $this->resolveRelatedField($model, $fieldPath);
-        return !$this->getMetadatas($resolved['class'])->isAssociationWithSingleJoinColumn($resolved['field']);
+        $metadata = $this->getMetadatas($resolved['class']);
+        return $metadata->hasAssociation($resolved['field']) && !$metadata->isAssociationWithSingleJoinColumn($resolved['field']);
     }
 
     /**
--- a/Guesser/PropelORMFieldGuesser.php
+++ b/Guesser/PropelORMFieldGuesser.php
@@ -73,6 +73,8 @@ class PropelORMFieldGuesser
         if ($relation) {
             return \RelationMap::MANY_TO_MANY === $relation->getType();
         }
+
+        return false;
     }
 
     /**


Resources/templates/CommonAdmin/generic_actions.php.twig
```diff
@@ -1,42 +1,42 @@
...
-                    {% endif %}
-
-                    {% if action.name is same as('excel') and excelActions|length %}
+    {% if builder.actions|default is not empty %}
+        {{ echo_block('pre_generic_actions') }}{{ echo_endblock() }}
+        {% set actionCredentials = builder.actions|mapBy('credentials')|flatten|clean|unique %}
+
+        {% if actionCredentials is not empty %}
+            {{ echo_set('actionCredentials', echo_twig_arr(actionCredentials), false) }}
+            {{ echo_if('is_one_admingenerator_granted(actionCredentials, (' ~ builder.ModelClass ~  ' is defined) ? ' ~ builder.ModelClass ~  ' : null )') }}
```
TheCelavi: Here is one change, note ternary operator? Second argument is passed to this function, or null.

Resources/templates/CommonAdmin/generic_actions.php.twig
```diff
@@ -1,42 +1,42 @@
...
+        {% endif %}
+
+        {% set excelActions = builder.excelActions is defined ? builder.excelActions : [] %}
+        {% for action in builder.Actions %}
+            {{ echo_block("generic_action_" ~ action.twigName) }}
+            {% if action.credentials %}
+                {{ echo_if_granted(action.credentials, builder.ModelClass) }}
```
TheCelavi: Here is change. echo_if_granted function generates proper ternary operator, so no need for manually generating expression.

TheCelavi:
Note that this is a bugfix, no BC break. Who ever had a issue with this, probably overrided problematic block. 

@bobvandevijver Dude, I can not express how desperate I need this fix :) Thank you. Do yell if you need additional clarification and/or output example.

bobvandevijver:
@TheCelavi This PR is not clear due to the massive amount of changes which are due to indentation changes. As you only have one file, can you resubmit only the changes you made? It should not be too hard :+1:

PR #300, opened 2016-07-26
Fix for generator of generic actions, should provide object to context

TheCelavi:
Object should be provided to context of access decision manager / jms security extra, if available, explained here: https://github.com/symfony2admingenerator/GeneratorBundle/pull/299

Cleaned up PR.


bobvandevijver:
Merged, thanks!

TheCelavi:
Thank you!

PR #313, opened 2017-08-21
Fixed #312, UUID can be used for entity IDs.

### LINKED ISSUE #312: Bad assumption based on entity ID in form templates when generating routes.
Generated code for new form and edit form in twig templates is:

    {% if User.id %}
        {% set attr = {'class': 'form-model-update', 'action': updateUrl}|merge(form.vars.attr|default({})) %}
    {% else %}
         {% set attr = {'class': 'form-model-new', 'action': createUrl}|merge(form.vars.attr|default({})) %}
    {% endif %}

Assumption is based that ID of entity will not exists if entity is now. However, if UUID is used and pre-generated as optimization method (and good practice in general), this fails.

`updateUrl` and `createUrl` are provided from controllers, those are just enough to check for existence in order to generate proper route.

TheCelavi:
Fix for https://github.com/symfony2admingenerator/GeneratorBundle/issues/312.

Controllers will send either `updateUrl` or `createUrl`, so that is reliable source which controller is being used (edit or new) in order to generate appropriate route.

In that matter, pre-generated entity ID (such as UUID) can be used safely. 

NOTE: This PR does not introduces BC breaks.

### DIFF
--- a/Resources/templates/CommonAdmin/EditTemplate/EditBuilderTemplate.php.twig
+++ b/Resources/templates/CommonAdmin/EditTemplate/EditBuilderTemplate.php.twig
@@ -23,7 +23,7 @@
 {{ echo_block("page_content") }}
     <div class="row">
         <div class="form-model form-model-{{ builder.modelClass|lower }} col-md-12">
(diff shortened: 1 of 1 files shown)


tobias-93:
@TheCelavi Thank you!

PR #315, opened 2017-09-08
Prevent double translation in Filter form label by treating it the same as edit or new form labels

tobias-93:
Filter form labels are translated by default, which could lead to double translations if the project also implements translation of form labels.
However, edit and new forms do not do this translation, leading to inconsistent behavior.

This PR proposes a change in the build-up of form rows in the filter form, making the behavior for new, edit and filter forms the same.
It seems that some 'functionality' is lost (being able to give the filter form row a different label than the other forms), but this can easily be achieved by passing the label in the filterOptions or by changing it in the list builder.
Therefore, no real functionality is lost.


sescandell:
👍

PR #316, opened 2017-11-11
Missing var added for variable declaration in scope

coke54:
This missing declaration is braking js


bobvandevijver:
Merged, thanks!

PR #323, opened 2019-04-30
Hotfix/sizeof not compatible with php7.2

TheCelavi:
sizeof() throws an exception in PHP 7.2+. This is quick fix for the issue.

### DIFF
--- a/Builder/Admin/BaseBuilder.php
+++ b/Builder/Admin/BaseBuilder.php
@@ -249,10 +249,10 @@ class BaseBuilder extends GenericBaseBuilder
         $display = $this->getVariable('display');
 
         // tabs
-        if (null == $display || 0 == sizeof($display)) {
+        if (null == $display || (is_array($display) && 0 == count($display))) {
             $tabs = $this->getVariable('tabs');
 
-            if (null != $tabs || 0 < sizeof($tabs)) {
+            if (null != $tabs || (is_array($tabs) && 0 < count($tabs))) {
                 $display = array();
 
                 foreach ($tabs as $tab) {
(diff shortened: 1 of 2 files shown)


TheCelavi:
@cedriclombardot @bobvandevijver and others - Ping. Sorry to bother you, but I have to express my concerns about this project. Even though there is no too much activities on it and it seams stalled, we have few production applications using this bundle - therefore, we are quite interested in hotfixes/security patches. In that matter, timely implementation and release of such fixes is kinda important for us, we don't like the idea to use our fork in project.

In that matter - if I can help somehow with code review, merging, release, etc... please let me know, it would be my pleasure. Thanks! Sorry again for ping and being annoying :)

bobvandevijver:
I will take a look at it tonight, because I want to test this change first. And tonight I hopefully have some time to do so :)

bobvandevijver:
@TheCelavi  Do you now what is in the variables that causes the warning? Might it be due to a wrong configuration on your side? As it seems it should always be either null (hence the first case), or an array.

Anyways, this seems to work fine, so I'll merge it. But I'm still curious what causes the error on your side!

TheCelavi:
I have checked all configuration files in regards to "display" key, literally - everything.

There are null values:
display: ~

There are arrays:

display: [some, fields]
display:
     - some
     - fields

and associative arrays

display:
    section: [some, fields]

Nothing seams out of the usual and not according to documentation. Everything worked on PHP 7.1, upgrade to 7.2 broke the damn thing...

VadimBondarenko:
Hi. I have the same problem with PHP 7.2.

bobvandevijver:
@VadimBondarenko v2.5.2 contains this fix, so that version should resolve your issue.

PR #327, opened 2023-08-09
Prevent 404 when requesting an out of range page

bobvandevijver:
This will allow PagerFanta to normalize the requested page if for whatever reason the page no longer exists in the admin list, for example after a bulk delete action on the last page.


tobias-93:
Makes one wonder why this isn't the default, considering the drawbacks of not normalizing... Anyway, thanks @bobvandevijver!

PR #328, opened 2023-10-26
Add option to generate base files outside cache directory

bobvandevijver:
This brings several benefits such as IDE completion and it resolves autoload/container compile issues when the cache is stale. You are required to run the `bin/console admin:generate-base-classes` to update the files after each configuration adjustment.


Resources/doc/install/general-configuration.md
```diff
@@ -52,6 +54,23 @@ admingenerator_generator:
...
 ```
 
+### Generate in project dir
+
+`generate_base_in_project_dir`: __default__: `false` __type: `boolean`
+`generate_base_in_project_dir_directory`: __default: `admin` __type: `string`
+
+Normally, the admingenerator generates the base files in the cache directory, which can have a negative impact on the
+IDE DX and requires adjustments to autoload the generator classes correctly. If set to true, the admin generator will
+generate the base admin files in `%kernel.project_dir%/src/Admingenerated` instead. You will need to run the
```
tobias-93, line 64: Default is probably `%kernel.project_dir%/admin/Admingenerated`?

tobias-93 (approved):

PR #334, opened 2024-03-12
Fix deprecations by adding return types and redefining the Command co…

tobias-93:
…nfiguration. Drop support for SF <5.4


Command/AssetsInstallCommand.php
```diff
@@ -3,25 +3,23 @@
...
 
 /**
- * Class AssetsInstallCommand
  * @package Admingenerator\GeneratorBundle\Command
  * @author Stéphane Escandell
  *
  * Automatically call bower install to properly import bowers components.
  * Push them to the web root directory
  */
+#[AsCommand('admin:assets-install', 'Fetch bower declared dependencies and push them into web root directory')]
```
bobvandevijver, line 20: Description needs to be in the configure, the name in the xml configuration where the tag is applied.

bobvandevijver: And then we can still support 5.4

bobvandevijver (changes requested):

bobvandevijver (approved):

PR #335, opened 2024-05-28
Fix compatibility with PHPSpreadsheet v2

tobias-93:


composer.json
```diff
@@ -68,7 +68,7 @@
         "symfony/maker-bundle": "To use the command for autogenerating configuration"
     },
     "conflict": {
-        "phpoffice/phpspreadsheet": ">=2"
+        "phpoffice/phpspreadsheet": "<1.23.0"
```
bobvandevijver, line 71: I think this is better, right?

```suggestion
        "phpoffice/phpspreadsheet": "<2"
```

bobvandevijver (approved):

composer.json
bobvandevijver: Ah, correct, they added the new API in that version!

PR #342, opened 2024-12-27
Add PHPUnit test action

tobias-93:


bobvandevijver (changes requested):
Looks like a couple of changes unrelated to the test action addition have been added (changes signatures outside the test directory, doctrine & pagerfanta version bumbs). Those need to be in a separate PR to be able to generate correct release notes.

(we should also update the `UPGRADE.MD` for those and previous changes).

tobias-93:
Depends on #344

bobvandevijver (approved):

PR #356, opened 2026-05-26
Fix deprecation in request->get()

tobias-93:


Resources/views/templates/CommonAdmin/ListAction/scopes.php.twig
```diff
@@ -2,7 +2,7 @@
     public function scopesAction(Request $request)
     {
         $this->request = $request;
-        $this->setScope($this->request->get('group'), $this->request->get('scope'));
+        $this->setScope($this->request->attributes->get('group'), $this->request->attributes->get('scope'));
```
bobvandevijver: This should actually use the `_route_params` attribute as those are the real values that were matched in the route. At this point in time only those values can be guaranteed to match with the route, unless they were explicitly adjusted by the user in the `_route_params` (which doesn't make sense). 'group' and 'scope' are just to generic as attributes and have a much higher probability to be overwritten by something else during the request lifetime.

Resources/views/templates/CommonAdmin/ListAction/filters.php.twig
```diff
@@ -10,8 +10,8 @@
     public function filtersAction(Request $request)
     {
         $this->request = $request;
-        if ($this->request->get('reset')) {
-            $this->setFilters(array());
+        if ($this->request->query->has('reset')) {
```
bobvandevijver: The reset button in the filter modal uses a posted form, so this change would break that button. Both query and post content will need to be checked.

bobvandevijver (changes requested):

bobvandevijver (approved):

